PR #12, opened 2019-12-02
ToString FromString Complete

angryjinyan:
Traits<std::chrono::time_point> ToString and FromString for ISO8601 format complete

### DIFF
--- a/src/Traits.h
+++ b/src/Traits.h
@@ -1011,36 +1011,46 @@ struct Traits<std::chrono::time_point<ClockT, DurationT>> : public TraitsImpl<st
     }
 
     static std::chrono::time_point<ClockT, DurationT> FromString(std::string const &value) {
-        std::string dash = "";
-        std::string colon = "";
-        std::string T = " ";
-        std::string Z = "Z";
-        std::string v(value);
+        // valid iso8601 format can be YYYY-MM-DDTHH:mm:ssZ
+        // dashes, T, colons and time zone offset can be optional
+        // so we search for these tokens and create a template to be used in parse
+        std::string dash_template_component = "";
+        std::string colon_template_component = "";
+        std::string T_template_component = " ";
+        std::string Z_template_component = "Z";
+
+        // use a pointer to make a copy only when necessary
+        std::string const * ptr(&value);
+        std::string modified;
+
         if ((value.find_first_of("Z") == std::string::npos)) {
             // there is no Z in the string
             if (value.find_last_of(":") == value.size() - 3) {
                 // there is a colon in the time zone
                 // remove it
-                v.erase(v.size() - 3, 1);
+                modified = *ptr;
+                modified.erase(modified.size() - 3, 1);
+                ptr = &modified;
             }
-            Z = "%z";
+            Z_template_component = "%z";
         }
         if (!(value.find_first_of("-") == std::string::npos)) {
             // there is - in the string
-            dash = "-";
+            dash_template_component = "-";
         }
         if (!(value.find_first_of(":") == std::string::npos)) {
             // there is : in the string
-            colon = ":";
+            colon_template_component = ":";
         }
         if (!(value.find_first_of("T") == std::string::npos)) {
             // there is T in the string
-            T = "T";
+            T_template_component = "T";
         }
-        std::string date_template = "%Y" + dash + "%m" + dash + "%d" + T + "%H" + colon +"%M" + colon + "%S" + Z;
+        std::istringstream ss(*ptr);
+        std::ostringstream date_template;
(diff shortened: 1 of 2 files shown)


src/Traits.h
```diff
@@ -1007,16 +1007,45 @@ struct Traits<std::chrono::system_clock::duration> : public Impl::CommonDuration
...
+        }
+        if (!(value.find_first_of(":") == std::string::npos)) {
+            // there is : in the string
+            colon = ":";
+        }
+        if (!(value.find_first_of("T") == std::string::npos)) {
+            // there is T in the string
+            T = "T";
+        }
+        std::string date_template = "%Y" + dash + "%m" + dash + "%d" + T + "%H" + colon +"%M" + colon + "%S" + Z;
```
davidbrownellWork: Each of these calls to the + operator creates a new string. Consider using std::ostringstream to format this string, and then call .str() on that object once the formatting is complete.

src/Traits.h
```diff
@@ -1007,16 +1007,45 @@ struct Traits<std::chrono::system_clock::duration> : public Impl::CommonDuration
...
-        date::operator <<(out, tp);
-        out.flush();
-
-        return out.str();
+        return date::format("%FT%TZ", date::floor<std::chrono::seconds>(tp));
     }
 
     static std::chrono::time_point<ClockT, DurationT> FromString(std::string const &value) {
-        std::ignore = value; throw std::logic_error("Not Implemented Yet");
+        std::string dash = "";
```
davidbrownellWork: I don't believe that these values will be necessary once you use std::ostringstream.

src/Traits.h
```diff
@@ -1007,16 +1007,45 @@ struct Traits<std::chrono::system_clock::duration> : public Impl::CommonDuration
...
+        std::string colon = "";
+        std::string T = " ";
+        std::string Z = "Z";
+        std::string v(value);
```
davidbrownellWork: Is this copy necessary?

src/Traits.h
davidbrownellWork: Now that I read more of the code, I see that you are using these values to populate the date_template. Update these names to indicate that this is how they will be used "dash_template_component", "colon_template_component", etc.

src/Traits.h
```diff
@@ -1007,16 +1007,45 @@ struct Traits<std::chrono::system_clock::duration> : public Impl::CommonDuration
...
+        if ((value.find_first_of("Z") == std::string::npos)) {
```
davidbrownellWork: It would be helpful to prefix this section of code with a comment that indicates that you are creating the date template based on values discovered in the incoming string.

src/UnitTests/Traits_UnitTest.cpp
```diff
@@ -259,6 +259,89 @@ bool SerializeOnly(T const &value) {
...
+    std::string tp_str = "1975-02-28T12:02:15Z";
+    std::chrono::system_clock::time_point tp = Traits<std::chrono::system_clock::time_point>::FromString(tp_str);
+    std::string tp_res = Traits<std::chrono::system_clock::time_point>::ToString(tp);
+    CHECK(tp_res == tp_str);
+
+    date::sys_days dp = date::floor<date::days>(tp);
+    date::year_month_day ymd = date::year_month_day{dp};
+    date::time_of_day<std::chrono::seconds> time = date::make_time(std::chrono::duration_cast<std::chrono::seconds>(tp-dp));
+
+    CHECK(ymd.year() == date::year(1975));
```
davidbrownellWork: Remove this code

src/UnitTests/Traits_UnitTest.cpp
```diff
@@ -259,6 +259,89 @@ bool SerializeOnly(T const &value) {
...
+    // without time offset
+    std::string tp_str = "1975-02-28T12:02:15Z";
+    std::chrono::system_clock::time_point tp = Traits<std::chrono::system_clock::time_point>::FromString(tp_str);
+    std::string tp_res = Traits<std::chrono::system_clock::time_point>::ToString(tp);
+    CHECK(tp_res == tp_str);
+
+    date::sys_days dp = date::floor<date::days>(tp);
+    date::year_month_day ymd = date::year_month_day{dp};
+    date::time_of_day<std::chrono::seconds> time = date::make_time(std::chrono::duration_cast<std::chrono::seconds>(tp-dp));
+
```
davidbrownellWork: Add test cases for every scenario in your code.

src/Traits.h
angryjinyan: I think so since it might need to remove the colon in the timezone and the input type is const string.

PR #14, opened 2019-12-03
DocumentStatisticsEstimator

wangyems:
Refactor of InverseDocumentFrequencyEstimator


src/Featurizers/Components/UnitTests/DocumentStatisticsEstimator_UnitTest.cpp
```diff
@@ -0,0 +1,158 @@
...
+#define CATCH_CONFIG_MAIN
+#include "catch.hpp"
+
+#include "../../../3rdParty/optional.h"
+#include "../../TestHelpers.h"
+#include "../DocumentStatisticsEstimator.h"
+
+namespace NS = Microsoft::Featurizer;
+
+// TODO: Test for DocumentStatisticsTrainingOnlyPolicy ctor - empty decorator
```
davidbrownellWork: Add this test or remove this comment.

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -0,0 +1,335 @@
...
+// |
+// |  Details::DocumentStatisticsTrainingOnlyPolicy
+// |
+// ----------------------------------------------------------------------
+inline Details::DocumentStatisticsTrainingOnlyPolicy::DocumentStatisticsTrainingOnlyPolicy(void) :
+    _totalNumDocuments(0) {
+}
+
+inline Details::DocumentStatisticsTrainingOnlyPolicy::DocumentStatisticsTrainingOnlyPolicy(StringDecorator decorator) :
+    //DocumentStatisticsTrainingOnlyPolicy(), this gives the error: an initializer for a delegating constructor must appear alone
```
davidbrownellWork: I forgot about this error. Go ahead and remove this comment.

src/Featurizers/Components/UnitTests/DocumentStatisticsEstimator_UnitTest.cpp
```diff
@@ -0,0 +1,158 @@
...
+
+TEST_CASE("string_idf") {
+    using FrequencyMap                         = NS::Featurizers::Components::DocumentStatisticsAnnotationData::FrequencyMap;
+
+    FrequencyMap const                         label({{"orange",3}, {"apple", 1}, {"peach", 3}, {"grape", 2}, {"banana",1}});
+    std::vector<std::vector<std::string>> const            inputBatches({{" orange  apple  apple peach  grape "},
+                                                                        {" grape orange     peach peach banana"},
+                                                                        {"orange orange peach   peach orange "}});
+
+    FrequencyMap const                         predict(Test(inputBatches));
```
davidbrownellWork: This code isn't actually predicting - I was looking for the calls to a Transformer's execut method. Instead, it represents the annotation data that was created. Perhaps rename all instances of "predict" to "annotaitonData" (or something similar).

davidbrownellWork: Similarily, perhaps rename "Test" to "Train".

src/Featurizers/Components/UnitTests/DocumentStatisticsEstimator_UnitTest.cpp
```diff
@@ -0,0 +1,158 @@
...
+    CHECK(predict == label);
+}
+
+TEST_CASE("string_idf_single_appearance") {
```
davidbrownellWork: I'm assuming that each of these tests represents different input scenarios? If so, validate all properties of the annotation data for each of the scenarios.

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -0,0 +1,334 @@
...
+    static constexpr char const * const     NameValue = DocumentStatisticsEstimatorName;
+
+    // ----------------------------------------------------------------------
+    // |
+    // |  Public Methods
+    // |
+    // ----------------------------------------------------------------------
+    DocumentStatisticsTrainingOnlyPolicy(void);
+    DocumentStatisticsTrainingOnlyPolicy(StringDecorator decorator);
+
```
michaelgsharp: Does this also need the `FEATURIZER_MOVE_CONSTRUCTOR_ONLY`?

wangyems: It seems like all TrainingOnlyPolicy class in Components miss it. Is it intentional? @davidbrownellWork

wangyems: I think it would better to refactor it in another PR if it's necessary. So I mark this conversation resolved at this time

davidbrownellWork: I agree with Ye. Technically speaking, the policy will never be copied. However, it is pretty subtle why it isn't required here but it is required just about everywhere else.

PR #15, opened 2019-12-03
Added in missing move constructor macro to various transformers

michaelgsharp:
Added in missing move constructor macros to various transformers.


michaelgsharp:
@davidbrownellWork Turns out the StringTransformer does have the archive constructor. A different error was spilling over and caused that error to be generated. So no changes are required on that one.

PR #16, opened 2019-12-03
Jin/statistics

angryjinyan:
Finished general statistics for all types.

For integer and numerical types, min, max, count, sum and average are calculated.

For all other types, min, max and count are supported.


src/Featurizers/Components/StatisticsMetricsEstimator.h
```diff
@@ -0,0 +1,391 @@
...
+class BasicStatisticalAnnotationData {
+public:
+    // ----------------------------------------------------------------------
+    // |
+    // |  Public Data
+    // |
+    // ----------------------------------------------------------------------
+    T             const                           Min;
+    T             const                           Max;
+    std::uint64_t  const                          Count;
```
davidbrownellWork: Can you remove one of the spaces after uint64_t since you went to the work to horizontally align the values.

src/Featurizers/Components/StatisticsMetricsEstimator.h
```diff
@@ -0,0 +1,391 @@
...
+///
+template <typename T>
+class StandardStatisticalAnnotaionData : public BasicStatisticalAnnotationData<T> {
+public:
+    // ----------------------------------------------------------------------
+    // |
+    // |  Public Data
+    // |
+    // ----------------------------------------------------------------------
+    long double   const                            Sum;
```
davidbrownellWork: I think that the Sum should be a type whose value is based on T. For example, Sum doesn't need to be a numeric type if T is an integer type.

src/Featurizers/Components/StatisticsMetricsEstimator.h
```diff
@@ -0,0 +1,391 @@
...
+
+    void fit(InputType const &input);
+    BasicStatisticalAnnotationData<T> complete_training(void);
+
+    // ----------------------------------------------------------------------
+    // |
+    // |  Public Data
+    // |
+    // ----------------------------------------------------------------------
+    InputType                                 _min;
```
davidbrownellWork: These shouldn't be public. Consider making them protected so StandardStatsTrainingOnlyPolicy has access to them.

src/Featurizers/Components/StatisticsMetricsEstimator.h
```diff
@@ -0,0 +1,391 @@
...
+// ----------------------------------------------------------------------
+// ----------------------------------------------------------------------
+
+// ----------------------------------------------------------------------
+// |
+// |  BasicStatisticalAnnotationData
+// |
+// ----------------------------------------------------------------------
+template <typename T>
+BasicStatisticalAnnotationData<T>::BasicStatisticalAnnotationData(T min, T max, std::uint64_t count) :
```
davidbrownellWork: Can count be 0 or is that an error condition?

src/Featurizers/Components/StatisticsMetricsEstimator.h
```diff
@@ -0,0 +1,391 @@
...
+}
+
+// ----------------------------------------------------------------------
+// |
+// |  StandardStatisticalAnnotaionData
+// |
+// ----------------------------------------------------------------------
+template <typename T>
+StandardStatisticalAnnotaionData<T>::StandardStatisticalAnnotaionData(long double sum, std::double_t average, T min, T max, std::uint64_t count) :
+    BasicStatisticalAnnotationData<T>(min, max, count),
```
davidbrownellWork: Move these values when calling the Basic ctor.

src/Featurizers/Components/StatisticsMetricsEstimator.h
```diff
@@ -0,0 +1,391 @@
...
+// |
+// ----------------------------------------------------------------------
+
+template <typename T>
+void update_basic_statistics(T input, T& min, T& max, std::uint64_t& count, bool& first_element_flag) {
+    // first non-null element will set min and max
+    if (first_element_flag) {
+        min = input;
+        max = input;
+        ++count;
```
davidbrownellWork: This logic is duplicated with the logic on 298. Right now, that isn't a big deal as it is only a single line. However, this logic will get more complex if another value/additional logic is added. Consider changing this to:

if(first_element_flag) {
    min = input;
    max = input;
    first_element_flag = false;
}
else {
    // Min / max logic here
}

++count;

src/Featurizers/Components/StatisticsMetricsEstimator.h
```diff
@@ -0,0 +1,391 @@
...
+    }
+    if(input > max) {
+        max = input;
+    }
+    // check if count will be out of bounds
+    if (std::numeric_limits<std::uint64_t>::max() == count) {
+        throw std::runtime_error("Overflow occured for count during calculating statistic metrics! Check your data!");
+    }
+    ++count;
+    return;
```
davidbrownellWork: This return statement isn't necessary.

src/Featurizers/Components/StatisticsMetricsEstimator.h
```diff
@@ -0,0 +1,391 @@
...
+    void fit(InputType const &input);
+    StandardStatisticalAnnotaionData<T> complete_training(void);
+private:
+    // ----------------------------------------------------------------------
+    // |
+    // |  Private Data
+    // |
+    // ----------------------------------------------------------------------
+    long double                               _sum;
+    std::double_t                             _average;
```
davidbrownellWork: This value doesn't need to be a member, as it is only used when creating the annotation data.

src/Featurizers/Components/StatisticsMetricsEstimator.h
```diff
@@ -0,0 +1,391 @@
...
+// ----------------------------------------------------------------------
+// |
+// |  Details::StandardStatsTrainingOnlyPolicy
+// |
+// ----------------------------------------------------------------------
+template <typename T>
+Details::StandardStatsTrainingOnlyPolicy<T>::StandardStatsTrainingOnlyPolicy(void) :
+    _sum(0),
+    _average(0) {
+        BasicStatsTrainingOnlyPolicy<T>::_min = 0;
```
davidbrownellWork: Are these statements necessary? Won't initialization be automatically handled in the first call to update_basic_statistics.

src/Featurizers/Components/StatisticsMetricsEstimator.h
```diff
@@ -0,0 +1,391 @@
...
+#endif
+
+// ----------------------------------------------------------------------
+// |
+// |  update functions
+// |
+// ----------------------------------------------------------------------
+
+template <typename T>
+void update_basic_statistics(T input, T& min, T& max, std::uint64_t& count, bool& first_element_flag) {
```
davidbrownellWork: This function feels like it should be a class. It is unusual to pass in 4 in/out parameters.

src/Featurizers/Components/UnitTests/StatisticsMetricsEstimator_UnitTest.cpp
```diff
@@ -0,0 +1,228 @@
...
+// ----------------------------------------------------------------------
+#define CATCH_CONFIG_MAIN
+#include "catch.hpp"
+
+#include "../../../3rdParty/optional.h"
+#include "../../TestHelpers.h"
+#include "../StatisticsMetricsEstimator.h"
+
+namespace NS = Microsoft::Featurizer;
+
```
davidbrownellWork: Add tests for any constructor that throws.

src/Featurizers/Components/StatisticsMetricsEstimator.h
```diff
@@ -0,0 +1,515 @@
...
+namespace Featurizer {
+namespace Featurizers {
+namespace Components {
+
+static constexpr char const * const         StatisticalMetricsEstimatorName("StatisticalMetricsEstimator");
+
+namespace TypeSelector {
+
+template <typename T>
+struct SumTypeSelector {
```
davidbrownellWork: Add comment

src/Featurizers/Components/StatisticsMetricsEstimator.h
```diff
@@ -0,0 +1,515 @@
...
+    // ----------------------------------------------------------------------
+    // |
+    // |  Public Methods
+    // |
+    // ----------------------------------------------------------------------
+    StandardStatsUpdater(void);
+
+    void update(InputType input);
+    // commit returns the result of min, max, count, sum and average
+    std::tuple<typename TypeSelector::SumTypeSelector<T>::type, std::double_t, InputType, InputType, std::uint64_t> commit(void);
```
davidbrownellWork: Return a struct

src/Featurizers/Components/StatisticsMetricsEstimator.h
```diff
@@ -0,0 +1,515 @@
...
+    // |  Public Methods
+    // |
+    // ----------------------------------------------------------------------
+    // first_element_flag is default to be true in constructor to guarantee
+    // min, max and count are set based on first valid input element
+    BasicStatsUpdater(void);
+
+    void update(InputType input);
+    // commit returns the result of min, max and count
+    std::tuple<InputType, InputType, std::uint64_t> commit(void);
```
davidbrownellWork: Return a struct

src/Featurizers/Components/StatisticsMetricsEstimator.h
```diff
@@ -0,0 +1,515 @@
...
+protected:
```
davidbrownellWork: Make this private.

src/Featurizers/Components/UnitTests/StatisticsMetricsEstimator_UnitTest.cpp
```diff
@@ -0,0 +1,235 @@
...
+#if (defined __clang__)
+#   pragma clang diagnostic push
+#   pragma clang diagnostic ignored "-Wfloat-equal"
+#   pragma clang diagnostic ignored "-Wdouble-promotion"
+#endif
+// test for invalid argument for constructors
+TEST_CASE("invalid argument for constructors") {
+    CHECK_THROWS_WITH(NS::Featurizers::Components::BasicStatisticalAnnotationData<int>(10,1,3), "min is > max");
+    CHECK_THROWS_WITH(NS::Featurizers::Components::StandardStatisticalAnnotaionData<int>(100, 17, 18, 30, 3), "average is not in the correct range");
+}
```
davidbrownellWork: Add a test for the other error condition.

src/Featurizers/Components/StatisticsMetricsEstimator.h
angryjinyan: I feel like I can directly return AnnotationData for basic and standard since a struct would be exactly the same as AnnotationData.

davidbrownellWork: I would recommend implementing AnnotationData in terms of these new structs.

PR #17, opened 2019-12-03
Fixed CatImputer to only take 1 template type instead of 2.

michaelgsharp:
Fixed CatImputer to only take 1 template type instead of 2. This change applies to the CatImputerEstimatorImpl and the CatImputerTransformer.

### DIFF
--- a/src/Featurizers/CatImputerFeaturizer.h
+++ b/src/Featurizers/CatImputerFeaturizer.h
@@ -17,6 +17,21 @@ namespace Featurizer {
 namespace Featurizers {
 
 /////////////////////////////////////////////////////////////////////////
+///  \class         CatImputerTraits
+///  \brief         Traits for mapping the input/output types for the CatImptuter.
+///                 This allows us to only change one place if typings need to change.
+///
+
+template <typename TransformedT>
+class CatImputerTraits{
+    using InputType = Traits<TransformedT>::nullable_type;
+    static_assert(Traits<InputType>::IsNullableType, "'InputT' must be a nullable type");
+    static_assert(Traits<TransformedT>::IsNullableType == false || Traits<TransformedT>::IsNativeNullableType, "'TransformedT' must not be a nullable type");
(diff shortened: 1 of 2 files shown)


src/Featurizers/CatImputerFeaturizer.h
```diff
@@ -222,9 +234,9 @@ CatImputerEstimator<TransformedT, MaxNumTrainingItemsV>::CatImputerEstimator(Ann
     BaseType(
         "CatImputerEstimator",
         pAllColumnAnnotations,
-        [pAllColumnAnnotations, colIndex](void) { return Components::HistogramEstimator<typename Traits<TransformedT>::nullable_type, MaxNumTrainingItemsV>(std::move(pAllColumnAnnotations), std::move(colIndex)); },
-        [pAllColumnAnnotations, colIndex](void) { return Components::ModeEstimator<typename Traits<TransformedT>::nullable_type, false, MaxNumTrainingItemsV>(std::move(pAllColumnAnnotations), std::move(colIndex)); },
-        [pAllColumnAnnotations, colIndex](void) { return Details::CatImputerEstimatorImpl<typename Traits<TransformedT>::nullable_type, TransformedT, MaxNumTrainingItemsV>(std::move(pAllColumnAnnotations), std::move(colIndex)); }
+        [pAllColumnAnnotations, colIndex](void) { return Components::HistogramEstimator<typename CatImputerTraits<TransformedT>::InputType, MaxNumTrainingItemsV>(std::move(pAllColumnAnnotations), std::move(colIndex)); },
```
davidbrownellWork: Can we use InputType here like what is being done on line 218?

michaelgsharp: `CatImputerEstimator` doesn't have the same `InputType` using statement that the `CatImputerTransformer` does. I can add it in if you would like these to use `InputType` here, but otherwise we have to leave it as is.

PR #18, opened 2019-12-04
add Eigen library

wangyems:
add Eigen submodule to 3rd party


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "src/3rdParty/eigen-git-mirror"]
```
davidbrownellWork: Let's tie this to a specific commit.

wangyems: Sure

PR #19, opened 2019-12-05
Adding GrainEstimatorImpl and unit tests

davidbrownellWork:
This Estimator wrapper maintains grain-specific (aka unique key) estimators and transformers.

### DIFF
--- a/src/Featurizers/Components/Details/EstimatorTraits.h
+++ b/src/Featurizers/Components/Details/EstimatorTraits.h
@@ -15,7 +15,7 @@ namespace Details {
 ///  \brief         Base declaration to determine if a type has the method
 ///                 'has_created_transformer'.
 ///
-template <typename, typename T>
+template <typename T, typename FunctionT>
 struct HasHasCreatedTransformerMethodImpl {
     static_assert(std::integral_constant<T, false>::value, "Second template parameter must be a function type");
 };
--- a/src/Featurizers/Components/Details/EstimatorTraits.h
+++ b/src/Featurizers/Components/Details/EstimatorTraits.h
@@ -1,0 +1,105 @@
+// ----------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License
+// ----------------------------------------------------------------------
+#pragma once
+
+namespace Microsoft {
+namespace Featurizer {
+namespace Featurizers {
+namespace Components {
+namespace Details {
+
+/////////////////////////////////////////////////////////////////////////
+///  \class         HasHasCreatedTransformerMethodImpl
+///  \brief         Base declaration to determine if a type has the method
+///                 'has_created_transformer'.
+///
+template <typename, typename T>
+struct HasHasCreatedTransformerMethodImpl {
+    static_assert(std::integral_constant<T, false>::value, "Second template parameter must be a function type");
+};
+
+/////////////////////////////////////////////////////////////////////////
+///  \class         HasHasCreatedTransformerMethodImpl
+///  \brief         Partial template specialization that is able to detect
+///                 a method's return type and arguments.
+///
+template <typename T, typename ReturnT, typename... ArgTs>
+class HasHasCreatedTransformerMethodImpl<T, ReturnT (ArgTs...)> {
+private:
+    template <typename U> static constexpr std::false_type Check(...);
+
+    template <typename U>
+    static constexpr std::true_type Check(
+        U *,
+        std::enable_if_t<
+            std::is_same<
+                decltype(std::declval<U>().has_created_transformer(std::declval<ArgTs>()...)),
+                ReturnT
+            >::value,
+            void *
+        >
+    );
+
+public:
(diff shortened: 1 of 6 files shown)


src/Featurizers/Components/Details/EstimatorTraits.h
```diff
@@ -0,0 +1,105 @@
...
+namespace Featurizers {
+namespace Components {
+namespace Details {
+
+/////////////////////////////////////////////////////////////////////////
+///  \class         HasHasCreatedTransformerMethodImpl
+///  \brief         Base declaration to determine if a type has the method
+///                 'has_created_transformer'.
+///
+template <typename, typename T>
```
michaelgsharp: The first `typename` doesn't have an identifier (is that the right word) with it. Was that intended? If so, what is the behavior? I haven't seen that before.

src/Featurizers/Components/UnitTests/CMakeLists.txt
```diff
@@ -42,6 +44,31 @@ foreach(_test_name IN ITEMS
...
+#
+# disable that test for now.
+
+if(
+    WIN32                                                                       # 1
+    AND "$ENV{DEVELOPMENT_ENVIRONMENT_CPP_COMPILER_NAME}" STREQUAL "Clang-8"    # 2
+    AND "$ENV{DEVELOPMENT_ENVIRONMENT_CPP_ARCHITECTURE}" STREQUAL "x86"         # 3
+    AND "${CMAKE_BUILD_TYPE}" MATCHES "^Release.*$"                             # 4
+)
+    # TODO: Test this with Clang 8.0.1
```
michaelgsharp: Didn't you mention yesterday that the bug still happens with 8.0.1?

src/Featurizers/Components/Details/EstimatorTraits.h
davidbrownellWork: It is like a function parameter that isn't used, where the type of the parameter is defined but the name is not. In this case, the struct is used as a baseline for partial template specialization. I am going to update the code to make this a bit clearer, as there is too much subtlety in the way that it is written right now.

src/Featurizers/Components/UnitTests/CMakeLists.txt
davidbrownellWork: I did more research, and it happens with clang 8.0.0 and 8.0.1, but not with clang 9.0.0. I will update the comments accordingly.

src/Featurizers/Components/GrainEstimatorImpl.h
```diff
@@ -0,0 +1,642 @@
...
+    TransformerMap                          _transformers;
+
+    // ----------------------------------------------------------------------
+    // |
+    // |  Private Methods
+    // |
+    // ----------------------------------------------------------------------
+
+    // MSVC has problems when the declaration and definition are separated
+    void execute_impl(typename BaseType::InputType const &input, typename BaseType::CallbackFunction const &callback) override {
```
michaelgsharp: I figured out more info about why MSVC has issues with this. It has to do with how many times it has to "unwrap" a type to find the actual type. Im sure thats not really the root cause, but it seems like if it has to take more then about 3 steps to find the actual type it doesn't succeed. I was able to repro it on a much smaller scale. We can sync on it some time.

src/Featurizers/Components/GrainEstimatorImpl.h
```diff
@@ -0,0 +1,642 @@
...
+
+                // If here, something was added. We expect that Annotation to be associated with the grain-based Estimator.
+                AnnotationMap::iterator     iter(map.find(kvp.second.Name));
+
+                if(iter == map.end())
+                    throw std::runtime_error("Unexpected AnnotationMap insertion (different Estimator)");
+
+                // There should be only one entry
+                if(iter->second.size() != 1)
+                    throw std::runtime_error("Unexpected AnnotationMap size");
```
michaelgsharp: Can an estimator not add more then one annotation? Or if that happens is it probably because it wasn't decomposed enough?

davidbrownellWork: A templated featurizer class may lead to multiple concrete templates for a particular featurizer (for example, a featurizer configured to train on the first 10 items vs. a featurizer that is configured to train on all of the items). This check makes sure that an annotation from only 1 concrete template of this particular featurizer was added.

src/Featurizers/Components/Details/EstimatorTraits.h
```diff
@@ -15,7 +15,7 @@ namespace Details {
 ///  \brief         Base declaration to determine if a type has the method
 ///                 'has_created_transformer'.
 ///
-template <typename, typename T>
+template <typename T, typename FunctionT>
 struct HasHasCreatedTransformerMethodImpl {
     static_assert(std::integral_constant<T, false>::value, "Second template parameter must be a function type");
 };
```
michaelgsharp: Don't you also need to switch the `T` in the static assert too?

davidbrownellWork: Either one will work in this case, as it is just needs to be a template so that it will cause a compiler error but only be invoked if the specialization is selected for compilation.

PR #20, opened 2019-12-06
Add more features to DocumentStatsEstimator

wangyems:


src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -74,6 +77,21 @@ struct IterRangeComp {
...
+///  \struct        FrequencyAndIndexStruct
+///  \brief         This struct is a combination of values of FrequencyMap and 
+///                 IndexMap for the purpose of keys equivelence of two maps
+///
+struct FrequencyAndIndexStruct {
+    std::uint32_t const TermFrequency;                            // Words and the number of documents that it appears in
+    std::uint32_t const Index;                              
+
+    //will cause error if the struct marked deleted after copy
+    //FEATURIZER_MOVE_CONSTRUCTOR_ONLY(FrequencyAndIndexStruct);
```
davidbrownellWork: This shouldn't cause an error - there is probably something else going on.

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -88,26 +106,110 @@ class DocumentStatisticsAnnotationData {
     // ----------------------------------------------------------------------
     using FrequencyMap                      = std::unordered_map<std::string, std::uint32_t>;
```
davidbrownellWork: Are these 2 using statements necessary after introducing FrequencyAndIndexMap?

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -137,8 +240,9 @@ class DocumentStatisticsTrainingOnlyPolicy {
     // |  Public Methods
     // |
     // ----------------------------------------------------------------------
-    DocumentStatisticsTrainingOnlyPolicy(void);
-    DocumentStatisticsTrainingOnlyPolicy(StringDecorator decorator);
+    DocumentStatisticsTrainingOnlyPolicy(StringDecorator decorator, nonstd::optional<IndexMap> vocabulary, nonstd::optional<std::uint32_t> maxFeatures, std::float_t minDf, std::float_t maxDf);
```
davidbrownellWork: Can you change the name of vocabulary to indicate that this is the vocabulary produced as part of a previous training session? Maybe "existingVocabulary"?

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -240,25 +374,92 @@ inline DocumentStatisticsAnnotationData::DocumentStatisticsAnnotationData(Freque
...
+// ----------------------------------------------------------------------
+inline Details::DocumentStatisticsTrainingOnlyPolicy::DocumentStatisticsTrainingOnlyPolicy(StringDecorator decorator, nonstd::optional<IndexMap> vocabulary, 
+                                                                                           nonstd::optional<std::uint32_t> maxFeatures, std::float_t minDf, 
+                                                                                           std::float_t maxDf) :
     _stringDecoratorFunc(
         [&decorator](void) -> StringDecorator & {
-            if(!decorator)
-                throw std::invalid_argument("decorator");
-
+            //temporary mute this validate
```
davidbrownellWork: Should this code be restored?

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -240,25 +374,92 @@ inline DocumentStatisticsAnnotationData::DocumentStatisticsAnnotationData(Freque
...
 }
 
-inline Details::DocumentStatisticsTrainingOnlyPolicy::DocumentStatisticsTrainingOnlyPolicy(StringDecorator decorator) :
+// ----------------------------------------------------------------------
+// |
+// |  Details::DocumentStatisticsTrainingOnlyPolicy
+// |
+// ----------------------------------------------------------------------
+inline Details::DocumentStatisticsTrainingOnlyPolicy::DocumentStatisticsTrainingOnlyPolicy(StringDecorator decorator, nonstd::optional<IndexMap> vocabulary, 
+                                                                                           nonstd::optional<std::uint32_t> maxFeatures, std::float_t minDf,
```
davidbrownellWork: I forget how maxFeatures is used. Is the the max vocab size? max num documents? max number of documents to look at? Consider making this a more descriptive name.

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -240,25 +374,92 @@ inline DocumentStatisticsAnnotationData::DocumentStatisticsAnnotationData(Freque
...
+        std::move(
+            [&maxFeatures](void) -> nonstd::optional<std::uint32_t> & {
+                if(maxFeatures.has_value() && *maxFeatures == 0)
+                    throw std::invalid_argument("maxFeaturesLowerBound");
+
+                return maxFeatures;
+            }()
+        )
+    ),
+    _minDf(
```
davidbrownellWork: Are NaN values valid for _minDf and _maxDf?

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -290,8 +491,24 @@ inline void Details::DocumentStatisticsTrainingOnlyPolicy::fit(InputType const &
...
 
 inline DocumentStatisticsAnnotationData Details::DocumentStatisticsTrainingOnlyPolicy::complete_training(void) {
-    IndexMap                                termIndex(Microsoft::Featurizer::Featurizers::Components::CreateIndexMap<std::string>(_termFrequency, IndexMap()));
-    return DocumentStatisticsAnnotationData(std::move(_termFrequency), std::move(termIndex), std::move(_totalNumDocuments));
+
+    //prune map by maxDf and minDf
+    FrequencyMap prunedTermFreqByDf = PruneTermFreqMap(_termFrequency, _minDf, _maxDf, static_cast<std::float_t>(_totalNumDocuments));
+
+    //prune map by maxFeatures
+    FrequencyMap prunedTermFreqByMaxFeatures = PruneTermFreqMap(prunedTermFreqByDf, _maxFeatures);
```
davidbrownellWork: As written, there will be 2 simultaneous copies of the frequency map - rewrite this so we only have 1 copy of the map.

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -88,26 +106,110 @@ class DocumentStatisticsAnnotationData {
...
     // |
     // ----------------------------------------------------------------------
-    DocumentStatisticsAnnotationData(FrequencyMap termFreq, IndexMap termIndex, std::uint32_t totalNumDocuments);
+    DocumentStatisticsAnnotationData(FrequencyAndIndexMap termFrequencyAndIndex, std::uint32_t totalNumDocuments);
     ~DocumentStatisticsAnnotationData(void) = default;
 
     FEATURIZER_MOVE_CONSTRUCTOR_ONLY(DocumentStatisticsAnnotationData);
 };
 
+namespace {
```
davidbrownellWork: Move this namespace to the Implementation section, as I don't believe that this functionality is used in the class declarations.

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -88,26 +106,110 @@ class DocumentStatisticsAnnotationData {
...
+
+inline DocumentStatisticsAnnotationData::FrequencyMap PruneTermFreqMap(DocumentStatisticsAnnotationData::FrequencyMap const & prunedTermFreqByDf, 
+                                                                       std::uint32_t maxFeatures) {
+    std::vector<std::uint32_t> frequencyVector;
```
davidbrownellWork: Add the statement `frequencyVector.reserve(prunedTermFreqByDf.size());` so we don't reallocate the vector multiple times as it grows.

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -88,26 +106,110 @@ class DocumentStatisticsAnnotationData {
...
+    DocumentStatisticsAnnotationData(FrequencyAndIndexMap termFrequencyAndIndex, std::uint32_t totalNumDocuments);
     ~DocumentStatisticsAnnotationData(void) = default;
 
     FEATURIZER_MOVE_CONSTRUCTOR_ONLY(DocumentStatisticsAnnotationData);
 };
 
+namespace {
+
+inline DocumentStatisticsAnnotationData::FrequencyMap PruneTermFreqMap(DocumentStatisticsAnnotationData::FrequencyMap const & prunedTermFreqByDf, 
+                                                                       std::uint32_t maxFeatures) {
```
davidbrownellWork: Should we immediately bail if the size of the incoming FrequencyMap is already < maxFeatures?

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -88,26 +106,110 @@ class DocumentStatisticsAnnotationData {
...
+        if (termFrequencyPair.second == targetValue && remainingTargetValue > 0) {
+            prunedTermFreq.insert(termFrequencyPair);
+            --remainingTargetValue;
+        }
+    }
+
+    return prunedTermFreq;
+}
+
+inline DocumentStatisticsAnnotationData::FrequencyMap PruneTermFreqMap(DocumentStatisticsAnnotationData::FrequencyMap const & termFrequency,
```
davidbrownellWork: I'd suggest that we remove those values that aren't in the desired range so that we aren't building multiple maps.

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -88,26 +106,110 @@ class DocumentStatisticsAnnotationData {
...
-    DocumentStatisticsAnnotationData(FrequencyMap termFreq, IndexMap termIndex, std::uint32_t totalNumDocuments);
+    DocumentStatisticsAnnotationData(FrequencyAndIndexMap termFrequencyAndIndex, std::uint32_t totalNumDocuments);
     ~DocumentStatisticsAnnotationData(void) = default;
 
     FEATURIZER_MOVE_CONSTRUCTOR_ONLY(DocumentStatisticsAnnotationData);
 };
 
+namespace {
+
+inline DocumentStatisticsAnnotationData::FrequencyMap PruneTermFreqMap(DocumentStatisticsAnnotationData::FrequencyMap const & prunedTermFreqByDf,
```
davidbrownellWork: I'd suggest that we remove values rather than recreate a new map to conserve memory.

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -88,26 +106,110 @@ class DocumentStatisticsAnnotationData {
...
+
+    return prunedTermFreq;
+}
+
+inline DocumentStatisticsAnnotationData::FrequencyMap PruneTermFreqMap(DocumentStatisticsAnnotationData::FrequencyMap prunedTermFreqByDf, 
+                                                                       nonstd::optional<std::uint32_t> maxFeatures) {
+    if (!maxFeatures.has_value() || prunedTermFreqByDf.empty())
+        return prunedTermFreqByDf;
+
+    std::uint32_t maxFeaturesValue = std::move(*maxFeatures);
```
davidbrownellWork: std::uint32_t const & maxFeaturizerValue(*maxFeatures);

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -88,26 +106,110 @@ class DocumentStatisticsAnnotationData {
...
+        std::float_t const freq = termFrequencyPair.second / totalNumDocumentsFloat;
+
+        if (freq >= minDf && freq <= maxDf) 
+            prunedTermFreq.insert(termFrequencyPair);
+    }
+
+    return prunedTermFreq;
+}
+
+inline DocumentStatisticsAnnotationData::FrequencyMap PruneTermFreqMap(DocumentStatisticsAnnotationData::FrequencyMap prunedTermFreqByDf,
```
davidbrownellWork: This method is very close to the definition above. Are they both necessary? If so, move them so they are defined next to each other. Also, add a comment explaining why they are both necessary.

src/Featurizers/Components/DocumentStatisticsEstimator.h
wangyems: Looks there are not necessary in the AnnotationData class, removed.

src/Featurizers/Components/DocumentStatisticsEstimator.h
wangyems: sure

src/Featurizers/Components/DocumentStatisticsEstimator.h
wangyems: it represents a TermFrequencyMap will be trimmed to a new map with words that have "top maxFeatures" of values.
e.g. Let's have an TermFrequencyMap of {{a, 3}, {b, 5}, {c, 1}, {d, 2}, {e, 6}}, assuming maxFeatures = 2, the TermFrequencyMap will be trimmed to { {b, 5},{e, 6} }; if maxFeatures = 4, the TermFrequencyMap will be trimmed to {{a, 3}, {b, 5},{d, 2}, {e, 6}}.

I think "TopKTerms" would be a better one

src/Featurizers/Components/DocumentStatisticsEstimator.h
wangyems: We only allow these two parameters in the range of [0.0f, 1.0f]

src/Featurizers/Components/DocumentStatisticsEstimator.h
wangyems: Yes, This logic is done in outer Function(Wrapper)

src/Featurizers/Components/DocumentStatisticsEstimator.h
wangyems: Agree, I believe it's better to do in one function

src/Featurizers/Components/DocumentStatisticsEstimator.h
wangyems: Not at this moment, because decorator will be generated from strings.h which is not finished yet. It will be restored as decorators are ready in next PR

src/Featurizers/Components/DocumentStatisticsEstimator.h
wangyems: you're right, it looks like copy occurs in generating FrequencyAndIndexMap. Added a comment there

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -86,7 +86,7 @@ struct FrequencyAndIndexStruct {
     std::uint32_t const TermFrequency;                            // Words and the number of documents that it appears in
     std::uint32_t const Index;                              
 
-    //will cause error if the struct marked deleted after copy
+    //copy needed in generating FrequencyAndIndexMap
```
davidbrownellWork: Perhaps use the map's emplace method? If that doesn't work out, let me know which line has the problems.

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -124,7 +140,9 @@ class DocumentStatisticsTrainingOnlyPolicy {
     using InputType                         = std::string;
     using InputTypeConstIterator            = std::string::const_iterator;
     using StringDecorator                   = std::function<std::string (InputTypeConstIterator begin, InputTypeConstIterator end)>;
-    using IndexMap                          = DocumentStatisticsAnnotationData::IndexMap;
+    using IndexMap                          = std::unordered_map<std::string, std::uint32_t>;
+    using FrequencyMap                      = IndexMap;
```
davidbrownellWork: It looks like this can be a private type.

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -240,23 +298,182 @@ inline DocumentStatisticsAnnotationData::DocumentStatisticsAnnotationData(Freque
...
 
-inline Details::DocumentStatisticsTrainingOnlyPolicy::DocumentStatisticsTrainingOnlyPolicy(StringDecorator decorator) :
+// ----------------------------------------------------------------------
+// |
+// |  Anonymous namespace
+// |
+// ----------------------------------------------------------------------
+namespace {
+
+inline Details::DocumentStatisticsTrainingOnlyPolicy::FrequencyMap PruneTermFreqMap(Details::DocumentStatisticsTrainingOnlyPolicy::FrequencyMap & termFrequency,
```
davidbrownellWork: This is just a nit, but I would take `termFrequency` by value and then return the modified `termFrequency`; that way, the caller can determine if they want a copy of the updated map or would rather like to move the memory into the function.

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -240,23 +298,182 @@ inline DocumentStatisticsAnnotationData::DocumentStatisticsAnnotationData(Freque
...
+                                                                                    std::float_t minDf, 
+                                                                                    std::float_t maxDf, 
+                                                                                    std::float_t totalNumDocumentsFloat,
+                                                                                    nonstd::optional<std::uint32_t> topKTerms) {
+
+    using FrequencyMap = Details::DocumentStatisticsTrainingOnlyPolicy::FrequencyMap;
+    
+    //trim by minDf and maxDf
+    if (minDf > 0.0f || maxDf < 1.0f) {
+        for (FrequencyMap::iterator termFrequencyIter = termFrequency.begin(); termFrequencyIter != termFrequency.end();) {
```
davidbrownellWork: It is uncommon to have a for loop that doesn't increment the variable within the loop. I recommend that you convert this to a while loop.

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -240,23 +298,182 @@ inline DocumentStatisticsAnnotationData::DocumentStatisticsAnnotationData(Freque
...
-inline Details::DocumentStatisticsTrainingOnlyPolicy::DocumentStatisticsTrainingOnlyPolicy(StringDecorator decorator) :
+// ----------------------------------------------------------------------
+// |
+// |  Anonymous namespace
+// |
+// ----------------------------------------------------------------------
+namespace {
+
+inline Details::DocumentStatisticsTrainingOnlyPolicy::FrequencyMap PruneTermFreqMap(Details::DocumentStatisticsTrainingOnlyPolicy::FrequencyMap & termFrequency, 
+                                                                                    std::float_t minDf,
```
davidbrownellWork: Putting this functionality into a single function is much more readable - nice job!

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -240,23 +298,182 @@ inline DocumentStatisticsAnnotationData::DocumentStatisticsAnnotationData(Freque
...
+                     frequencyVector.begin() + static_cast<int>(topKTermsValue) - 1, 
+                     frequencyVector.end(), 
+                     std::greater<std::uint32_t>());
+
+    std::uint32_t targetValue = frequencyVector[topKTermsValue - 1];
+    long long remainingTargetValue = std::count(frequencyVector.begin(), 
+                                                frequencyVector.begin() + static_cast<int>(topKTermsValue),
+                                                targetValue);
+
+    for (FrequencyMap::iterator termFrequencyIter = termFrequency.begin(); termFrequencyIter != termFrequency.end();) {
```
davidbrownellWork: Similar to the comments above, I'd turn this into a while loop. Each time I loop at this loop, I have to refer back to the for statement once I see the manual increment on line 393 to make sure that there isn't a doulbe-increment going on.

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -240,23 +298,182 @@ inline DocumentStatisticsAnnotationData::DocumentStatisticsAnnotationData(Freque
...
+                       return kv.second; 
+                   });
+
+    //partial sort, ave O(n)
+    std::nth_element(frequencyVector.begin(), 
+                     frequencyVector.begin() + static_cast<int>(topKTermsValue) - 1, 
+                     frequencyVector.end(), 
+                     std::greater<std::uint32_t>());
+
+    std::uint32_t targetValue = frequencyVector[topKTermsValue - 1];
```
davidbrownellWork: I believe that the frequencyVector is an implementation detail used to calculate the remainingTargetValue; if so, we don't need to keep frequencyVector around once that value is calculated. Consider putting this calculation in a lambda so that the vector is destroyed once that lambda is finished.

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -240,23 +298,182 @@ inline DocumentStatisticsAnnotationData::DocumentStatisticsAnnotationData(Freque
...
+    //trim by topKTerms
+    if (!topKTerms.has_value() || termFrequency.empty())
+        return termFrequency;
+
+    std::uint32_t const & topKTermsValue(*topKTerms);
+
+    if (topKTermsValue > termFrequency.size())
+        throw std::invalid_argument("topKTermsUpperBound");
+
+    std::vector<std::uint32_t> frequencyVector;
```
davidbrownellWork: I'm concerned about this logic - I see how it works during a full retrain scenario, but not clear how it maintains values after multiple retrains. Here is the scenario that I am thinking about:

  1) Train, 100 terms are calculated (meaning ids 0-99 (or 1-100) are assigned)
  2) Incremental training with original 100 items, where 10 new words are discovered (meaning ids 0-109 (or 1-110) are assigned)
  3) User removes random words in the training set (for example, maybe those words aren't meaningful) <<<< This may be an invalid operation; if so, disregard the rest of this
  4) We retrain, and IndexMap will fill in the gaps (I just verified that there is a bug in IndexMap where it potentially assigns duplicate values in this scenario)

So, the problem is that we will remove some of the values in the incoming map when I believe that we should always treat those as more important than the new terms that we discovered during retraining.

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -240,23 +298,182 @@ inline DocumentStatisticsAnnotationData::DocumentStatisticsAnnotationData(Freque
...
+    long long remainingTargetValue = std::count(frequencyVector.begin(), 
+                                                frequencyVector.begin() + static_cast<int>(topKTermsValue),
+                                                targetValue);
+
+    for (FrequencyMap::iterator termFrequencyIter = termFrequency.begin(); termFrequencyIter != termFrequency.end();) {
+        if (termFrequencyIter->second < targetValue) {
+            termFrequencyIter = termFrequency.erase(termFrequencyIter);
+            continue;
+        } else if (termFrequencyIter->second == targetValue) {
+            if (remainingTargetValue-- <= 0) {
```
davidbrownellWork: This line is subtle, as the value becomes smaller regardless of the result of the condition (which happens to be OK since we only want this branch to execute when we have populated N items). Is there a way to rewrite this logic to more clearly express what is happening?

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -240,23 +298,182 @@ inline DocumentStatisticsAnnotationData::DocumentStatisticsAnnotationData(Freque
...
+    }
+
+    return termFrequency;
+}
+
+inline DocumentStatisticsAnnotationData::FrequencyAndIndexMap MergeTwoMapsWithSameKeys(Details::DocumentStatisticsTrainingOnlyPolicy::FrequencyMap const & termFrequency, 
+                                                                                       Details::DocumentStatisticsTrainingOnlyPolicy::IndexMap const & termIndex) {
+    DocumentStatisticsAnnotationData::FrequencyAndIndexMap termFrequencyAndIndex;
+    for (auto const & termFrequencyIter : termFrequency) {
+        termFrequencyAndIndex.insert(std::make_pair(termFrequencyIter.first, FrequencyAndIndexStruct(termFrequencyIter.second, termIndex.at(termFrequencyIter.first))));
```
davidbrownellWork: I'd like to see an assertion here that verifies termFrequencyIter.first is in termIndex.

src/Featurizers/Components/DocumentStatisticsEstimator.h
davidbrownellWork: I don't follow. I believe that a decorator is required from the class's perspective? If so, the caller should be the one providing the temporary implementation now and the one in strings.h later.

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -240,23 +298,182 @@ inline DocumentStatisticsAnnotationData::DocumentStatisticsAnnotationData(Freque
...
+        }
+    }
+
+    //trim by topKTerms
+    if (!topKTerms.has_value() || termFrequency.empty())
+        return termFrequency;
+
+    std::uint32_t const & topKTermsValue(*topKTerms);
+
+    if (topKTermsValue > termFrequency.size())
```
davidbrownellWork: Is this an error condition? It seems reasonable to specify this value as 1000 even when only 200 unique terms are discovered in the training data.

src/Featurizers/Components/UnitTests/DocumentStatisticsEstimator_UnitTest.cpp
```diff
@@ -6,8 +6,9 @@
 #include "catch.hpp"
```
davidbrownellWork: I'll add comments to this file once the interface has stabilized. Please make sure that all scenarios are covered in tests.

src/Featurizers/Components/DocumentStatisticsEstimator.h
wangyems: highly suspect line 407, but not 100% sure as the error message doesn't target any. And it looks emplace() doesn't work out here.

davidbrownellWork: Try to change 
```
termFrequencyAndIndex.insert(std::make_pair(termFrequencyIter.first, FrequencyAndIndexStruct(termFrequencyIter.second, termIndex.at(termFrequencyIter.first))));
    }
```
to
```
FrequencyAndIndexStruct value(termFrequencyIter.second, termIndex.at(termFrequencyIter.first));
termFrequenceAndIndex.emplace(std::make_pair(termFrequencyIter.first, std::move(value)));
```

If that doesn't work:
```
FrequencyAndIndexStruct value(termFrequencyIter.second, termIndex.at(termFrequencyIter.first));
termFrequenceAndIndex.emplace(<The Map Type Here>::value_type(termFrequencyIter.first, std::move(value)));
```

src/Featurizers/Components/DocumentStatisticsEstimator.h
wangyems: FrequencyAndIndexMap can be a private type, while the functions in anonymous space uses the public IndexMap and FrequencyMap. Would it be better to move the functions from anonymous space to DocumentStatisticsTrainingOnlyPolicy? or just use the std::unordered_map as the return type?

src/Featurizers/Components/DocumentStatisticsEstimator.h
wangyems: looks both not work, working on trying other ways...

src/Featurizers/Components/DocumentStatisticsEstimator.h
wangyems: sure

src/Featurizers/Components/DocumentStatisticsEstimator.h
wangyems: get it~

src/Featurizers/Components/DocumentStatisticsEstimator.h
wangyems: I agree

src/Featurizers/Components/DocumentStatisticsEstimator.h
davidbrownellWork: If the type is only used in the anonymous namespace, it can be defined there.

PR #22, opened 2019-12-09
fixed memory leak in callback function

michaelgsharp:
fixed memory leak in callback function

### DIFF
--- a/src/Featurizer.h
+++ b/src/Featurizer.h
@@ -572,7 +572,12 @@ typename StandardTransformer<InputT, TransformedT>::TransformedType StandardTran
     );
 
     assert(isSet);
-    return std::move(*pResult);
+    
+    TransformedT                            result(std::move(*pResult));
+
+    pResult->~TransformedT();
+
+    return result;
 }
 
 template <typename InputT, typename TransformedT>


src/Featurizer.h
```diff
@@ -572,7 +572,12 @@ typename StandardTransformer<InputT, TransformedT>::TransformedType StandardTran
     );
 
     assert(isSet);
-    return std::move(*pResult);
+    
+    TransformedT                            result(std::move(*pResult));
+
+    pResult->~TransformedT();
+
+    return result;
```
yuslepukhin: I think this would work. Here is an alternative suggestion. Make the callback return an instance of TransformedType instead of void. It would construct the instance by value and return it. The compiler would move it. You will need to use a templated function and pass its address as a callback but you won't need a dynamica allocation and all that placement new and __dtor issues. But this would require some changes in the class hierarchy. For that reason, I can not do it.

src/Featurizer.h
```diff
@@ -572,7 +572,12 @@ typename StandardTransformer<InputT, TransformedT>::TransformedType StandardTran
     );
 
     assert(isSet);
-    return std::move(*pResult);
+    
+    TransformedT                            result(std::move(*pResult));
+
+    pResult->~TransformedT();
```
yuslepukhin: >pResult->~TransformedT(); [](start = 4, length = 25)

I did not see alloca.

src/Featurizer.h
davidbrownellWork: Unfortunately, this callback was introduced to handle scenarios where calls to `execute`  and `flush` returns 0-N items, so a single return value isn't viable (especially in the case of 0 return items).

Scenarios for 0 items:
  - `execute`: imputer with backfill strategy
  - `flush`: Most featurizers don't flush, so this is actually the common case

Scenario for N items:
  - `execute`: Featurizer that splits a sentence input string into words
  - `flush`: Featurizer that has queued multiple results for processing.

yuslepukhin: Good, it provides some more details I was not aware of. Is it possible then that alloca would attempt a large allocation on the stack?

---
In reply to: [355626053](https://github.com/microsoft/FeaturizersLibrary/pull/22#discussion_r355626053) [](ancestors = 355626053)

PR #25, opened 2019-12-10
Added Featurizers/CMakeLists.txt

davidbrownellWork:


src/Featurizers/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+# ----------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License
+# ----------------------------------------------------------------------
+cmake_minimum_required(VERSION 3.5.0)
+
+macro(Impl)
+    # Defining a function here to introduce a new scope for local variables
+    get_filename_component(_this_path ${CMAKE_CURRENT_LIST_FILE} DIRECTORY)
```
michaelgsharp: I'm just curious as I don't know cmake well enough, but where does the `${CMAKE_CURRENT_LIST_FILE}` come from?

src/Featurizers/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+
+    include(${_this_path}/cmake/FeaturizersCode.cmake)
```
michaelgsharp: On line 9 you have `_this_path`, but on line 11 you have `${_this_path}`, was that intended?

src/Featurizers/CMakeLists.txt
davidbrownellWork: [cmake provides it](https://cmake.org/cmake/help/latest/variable/CMAKE_CURRENT_LIST_FILE.html).

src/Featurizers/CMakeLists.txt
yuslepukhin: >${_this_path}/ [](start = 12, length = 14)

Can we remove ${_this_path}/  and just do cmake/Featurizers.code ? Anyone can read it.

davidbrownellWork: cmake syntax is awful - the usage is correct.

yuslepukhin: Let's make it readable

---
In reply to: [356252729](https://github.com/microsoft/FeaturizersLibrary/pull/25#discussion_r356252729) [](ancestors = 356252729)

PR #26, opened 2019-12-10
standard deviation and norm updater

angryjinyan:
Complete updater for standard deviation and norms

STDUpdater uses the NormUpdater to calculate standard deviation

Most of the checks are implemented within NormUpdater


src/Featurizers/Components/StandardDeviationUpdater.h
```diff
@@ -0,0 +1,148 @@
...
+    _updater(NormUpdater<std::double_t>()) {
+        // input type can only be integer or numeric
+        static_assert(Traits<T>::IsIntOrNumeric::value, "Input type to standard deviation updater has to be integer or numerical types!");
+        if (_count == 0) {
+            throw std::invalid_argument("Count is 0 when constructing STDUpdater to calculate standard deviation!");
+        }
+}
+
+template <typename T>
+void Details::STDUpdater<T>::update(T input) {
```
davidbrownellWork: Verify that internal coutner < _count (aka this function should only be called _count times)

src/Featurizers/Components/StandardDeviationUpdater.h
```diff
@@ -0,0 +1,148 @@
...
+    _updater.update_l2(static_cast<std::double_t>(input) - _average);
+}
+
+template <typename T>
+typename Details::STDUpdater<T>::STDResult Details::STDUpdater<T>::commit(void) {
+    long double const & l2(_updater.commit_l2().L2_norm);
```
davidbrownellWork: Verify that internal counter == _count (aka this function should only be called after _count calls to update)

src/Featurizers/Components/NormUpdaters.h
```diff
@@ -0,0 +1,413 @@
...
+
+}
+
+#if (defined __clang__)
+#   pragma clang diagnostic push
+#   pragma clang diagnostic ignored "-Wfloat-equal"
+#   pragma clang diagnostic ignored "-Wdouble-promotion"
+#endif
+
+namespace Details{
```
davidbrownellWork: These shouldn't be in the Details namespace, but probably should go into a "NormUpdaters" namespace. (just change the name)

src/Featurizers/Components/NormUpdaters.h
```diff
@@ -0,0 +1,413 @@
...
+///
+template <typename T>
+class L1NormUpdater {
+public:
+
+    /////////////////////////////////////////////////////////////////////////
+    ///  \class         L1NormResult
+    ///  \brief         A structure which contains the l1 norm
+    ///
+    struct L1NormResult {
```
davidbrownellWork: Nit: The struct name can be Result since it will always have to be referred to via its qualified name (`L1NormUpdater::L1NormResult` vs `LiNormUpdater::Result`).

src/Featurizers/Components/NormUpdaters.h
```diff
@@ -0,0 +1,413 @@
...
+
+    // ----------------------------------------------------------------------
+    // |
+    // |  Public Methods
+    // |
+    // ----------------------------------------------------------------------
+
+    L1NormUpdater(void);
+
+    void update_l1(InputType input);
```
davidbrownellWork: This can probably be `update`

src/Featurizers/Components/NormUpdaters.h
```diff
@@ -0,0 +1,413 @@
...
+
+    L1NormResult commit_l1(void);
```
davidbrownellWork: This can probably be `commit`

src/Featurizers/Components/NormUpdaters.h
```diff
@@ -0,0 +1,413 @@
...
+private:
+    // ----------------------------------------------------------------------
+    // |
+    // |  Private Data
+    // |
+    // ----------------------------------------------------------------------
+    long double _l1_norm;
+
+    // flags to check if update function is called
+    bool        _update_l1_flag;
```
davidbrownellWork: Are these update flags necessary? Is it an error to call commit without update?

src/Featurizers/Components/NormUpdaters.h
```diff
@@ -0,0 +1,413 @@
...
+// |
+// |  Details::L1NormUpdater
+// |
+// ----------------------------------------------------------------------
+template <typename T>
+Details::L1NormUpdater<T>::L1NormUpdater() :
+    _l1_norm(0),
+    _update_l1_flag(false) {
+        // input type can only be integer or numeric
+        static_assert(Traits<T>::IsIntOrNumeric::value, "Input type to norm updater has to be integer or numerical types!");
```
davidbrownellWork: static_asserts can appear anywhere in code, so it might be better to place these in the class declaration.

src/Featurizers/Components/NormUpdaters.h
davidbrownellWork: Now that I think about this, all of these updaters return a single value. Should this return a long double rather than a struct?

src/Featurizers/Components/NormUpdaters.h
angryjinyan: I think so. It makes no sense to just have a commit without update. So if this happens, an error is thrown to remind user about this.

src/Featurizers/Components/StandardDeviationUpdater.h
```diff
@@ -0,0 +1,156 @@
...
+namespace Featurizers {
+namespace Components {
+
+#if (defined __clang__)
+#   pragma clang diagnostic push
+#   pragma clang diagnostic ignored "-Wfloat-equal"
+#   pragma clang diagnostic ignored "-Wdouble-promotion"
+#endif
+
+namespace Details{
```
davidbrownellWork: This shouldn't be in a Details namespace, as it will be used by other code.

src/Featurizers/Components/StandardDeviationUpdater.h
```diff
@@ -0,0 +1,156 @@
...
+    // check if count will be out of bounds
+    if (std::numeric_limits<std::uint64_t>::max() == _counter) {
+        throw std::runtime_error("Overflow occured for count during calculating standard deviation! You might input too much data");
+    }
+    ++_counter;
+}
+
+template <typename T>
+typename Details::STDUpdater<T>::STDResult Details::STDUpdater<T>::commit(void) {
+    long double const l2(_updater.commit_l2().L2_norm);
```
davidbrownellWork: Throw if _counter == 0.

angryjinyan: I don't think this necessary since _updater.commit() will do the check to see if _counter is 0.

src/Featurizers/Components/StatisticsMetricsEstimator.h
```diff
@@ -166,8 +166,9 @@ class StandardStatsUpdater : public BasicStatsUpdater<T> {
     // ----------------------------------------------------------------------
     typename TypeSelector::SumTypeSelector<T>::type                               _sum;
 };
+}
```
davidbrownellWork: The name of the type in on 111 is different from the name on 116.

angryjinyan: Good catch! It's fixed now.

PR #28, opened 2019-12-11
Update for ORT code generation

michaelgsharp:
Updates to the ORT code generation to fix a couple of issues as well as changes based on feedback.


src/Tools/CodeGenerator/Plugins/OnnxRuntimePlugin.py
```diff
@@ -372,6 +373,34 @@ def CreateOutputStatement(
             """,
         ).format(index, output_name, output_documentation or "No information is available", output_type)
 
+    # ----------------------------------------------------------------------
+    def CreateTypeInferenceConstraints(output_type_mappings):
+        code = []
+        mapping_length = len(output_type_mappings)
```
davidbrownellWork: Looks like this isn't used.

src/Tools/CodeGenerator/Plugins/OnnxRuntimePlugin.py
```diff
@@ -372,6 +373,34 @@ def CreateOutputStatement(
...
+    def CreateTypeInferenceConstraints(output_type_mappings):
+        code = []
+        mapping_length = len(output_type_mappings)
+        constraint_format = "input_elem_type == ONNX_NAMESPACE::TensorProto_DataType_{input_type_upper}"
+
+        for index, (output_type, input_types) in enumerate(output_type_mappings.items()):
+            constraints = []
+            for input_type in input_types:
+                constraints.append(constraint_format.format(
+                    input_type_upper = re.sub(r'^std::|_t$', '', input_type).upper()
```
davidbrownellWork: Creating a regex is relatively expensive - it would be better to create it outside of the loop and then use it here.

src/Tools/CodeGenerator/Plugins/OnnxRuntimePlugin.py
```diff
@@ -428,7 +457,7 @@ def CreateMacro(
                                 type_name=type_name,
                                 constraints=", ".join(
                                     [
-                                        '"{}"'.format(constraint)
+                                        '"tensor({})"'.format(constraint[:-2] if constraint[-2:] == "_t" else constraint)
```
davidbrownellWork: Change this to `... if constraint.endswith("_t") else ...`, which will prevent errors if we happen to see a constraint with only 1 char.

src/Tools/CodeGenerator/Plugins/OnnxRuntimePlugin.py
michaelgsharp: I didn't do that because I read this in the documentation.
```
Note The compiled versions of the most recent patterns passed to re.compile() and the module-level matching functions are cached, so programs that use only a few regular expressions at a time needn’t worry about compiling regular expressions.
```
I am using `sub` not `match`, so will it still get cached in this case?

src/Tools/CodeGenerator/Plugins/OnnxRuntimePlugin.py
michaelgsharp: Good catch. I fixed that everywhere else but missed that one.

src/Tools/CodeGenerator/Plugins/OnnxRuntimePlugin.py
```diff
@@ -487,74 +516,107 @@ def CreateMacro(
...
-                if custom_struct_data is None or output_type not in custom_struct_data:
                     output_statements = [
                         CreateOutputStatement(
-                            "tensor({})".format(output_type.replace("std::", "")),
+                            "OutputT",
                             item.output_description,
                         ),
                     ]
 
+                if (len(all_output_types) == 1):
```
davidbrownellWork: Does it make sense to put the suffix creation code into the if/else statement above?

src/Tools/CodeGenerator/Plugins/OnnxRuntimePlugin.py
davidbrownellWork: Interesting - I didn't know that. My hunch is that it will cache for both sub and match.

src/Tools/CodeGenerator/Plugins/OnnxRuntimePlugin.py
michaelgsharp: Yeah. Initially there was only one suffix so it was below. I didn't think about moving it up when I added in the second one. I'll move it up.

src/Tools/CodeGenerator/Plugins/OnnxRuntimePlugin.py
michaelgsharp: Good catch. I have removed it.

michaelgsharp:
Closing and reopening to clear internal state.

PR #30, opened 2019-12-12
Fix up CentOS 7 build

yuslepukhin:


davidbrownellWork:
Most of our code is in header files, so building the library doesn't exercise those files. Running unit tests is a good way to ensure that the code will compile on a given platform and this is when you will start to run into problems with date.h and other 3rdParty files.

yuslepukhin:
Sure, but you don't build and run tests in ORT. And at this point ORT is fine. Should we advance to a later commit, I can address those.

Dmitri

________________________________

PR #33, opened 2019-12-12
Downgraded to C++11 in support of a wider variety of compilers

davidbrownellWork:


src/Featurizers/Components/Details/PipelineExecutionEstimatorImpl_details.h
```diff
@@ -865,7 +865,7 @@ class TransformerChainElement<
     }
 
     TransformerChainElement(Archive &ar) :
-        _pTransformer(std::make_unique<typename ThisEstimator::TransformerType>(ar)) {
+        _pTransformer(new typename ThisEstimator::TransformerType(ar)) {
```
michaelgsharp: Now that we are using a bunch of `new` instead of `std::make_unique` will we need to have a bunch of deletes as well? Or because it is moving will it still be ok?

davidbrownellWork: It is still being assigned to an unique_ptr, the change is that the make_unique function isn't available.

michaelgsharp:
Everything looks good to me. It seems like there was really only a couple of changes that were just propagated everywhere. How would would it be to enable the tests on the Holy Build Box and see if they pass now since some of the `emplace` were replaced and thats where you thought the issue was?

davidbrownellWork:
> 
> 
> Everything looks good to me. It seems like there was really only a couple of changes that were just propagated everywhere. How would would it be to enable the tests on the Holy Build Box and see if they pass now since some of the `emplace` were replaced and thats where you thought the issue was?

The problem on Holy Build Box wasn't the build but rather crashes on some tests. It's pretty easy to get that setup if you are interested in checking it out.

PR #37, opened 2019-12-16
StandardScaler

angryjinyan:
Finish StandardScalerFeaturizer using StatisticalMetricsEstimator and StandardDeviationEstimator in Components.

TODO: Add shared library interface for StandardScalerFeaturizer.


src/Featurizers/StandardScaleWrapperFeaturizer.h
```diff
@@ -0,0 +1,348 @@
...
+///
+template <typename InputT, typename TransformedT>
+class StandardScalerTransformer : public StandardTransformer<InputT, TransformedT> {
+public:
+    // ----------------------------------------------------------------------
+    // |
+    // |  Public Types
+    // |
+    // ----------------------------------------------------------------------
+    static_assert(Traits<TransformedT>::IsNullableType && Traits<TransformedT>::IsIntOrNumeric::value, "'TransformedT' must be a numeric type");
```
davidbrownellWork: Will TransformedT always be std::double? If not, it seems like it will always be a numeric type. This static_assert should assert that rather than if it is nullable. We should probably add a IsNumeric to Traits.h.

src/Featurizers/StandardScaleWrapperFeaturizer.h
```diff
@@ -0,0 +1,348 @@
...
+    // ----------------------------------------------------------------------
+    using BaseType                          = TransformerEstimator<InputT, TransformedT>;
+    using TransformerType                   = StandardScalerTransformer<InputT, TransformedT>;
+
+    // ----------------------------------------------------------------------
+    // |
+    // |  Public Methods
+    // |
+    // ----------------------------------------------------------------------
+    StandardScaleWrapperEstimatorImpl(AnnotationMapsPtr pAllColumnAnnotations, size_t colIndex, bool with_mean, bool with_std);
```
davidbrownellWork: Use consistent naming for parameters.

src/Featurizers/StandardScaleWrapperFeaturizer.h
```diff
@@ -0,0 +1,348 @@
...
+        using StatisticalMetricsEstimator       = Components::StatisticalMetricsEstimator<InputT, MaxNumTrainingItemsV>;
+
+        using StandardDeviationAnnotationData   = Components::StandardDeviationAnnotationData;
+        using StandardDeviationEstimator        = Components::StandardDeviationEstimator<InputT, MaxNumTrainingItemsV>;
+        // ----------------------------------------------------------------------
+        // if _with_mean is false, average is set to 0
+        std::double_t average = 0;
+        // if _with_std is false, deviation is set to 1
+        std::double_t deviation = 1;
+        StandardStatisticalAnnotationData const &        average_stats(StatisticalMetricsEstimator::get_annotation_data(this->get_column_annotations(), _colIndex, Components::StatisticalMetricsEstimatorName));
```
davidbrownellWork: These values should only be retrieved if the corresponding flags are set to true.

src/Featurizers/StandardScaleWrapperFeaturizer.h
```diff
@@ -0,0 +1,348 @@
...
+        }
+        assert(average_stats.Count == deviation_stats.Count);
+#if (defined __clang__)
+#   pragma clang diagnostic push
+#   pragma clang diagnostic ignored "-Wfloat-equal"
+#   pragma clang diagnostic ignored "-Wdouble-promotion"
+#endif
+        // check if deviation is 0
+        // if it's 0, change it to 1
+        if (deviation == 0) {
```
davidbrownellWork: Change this to 0.0 and 1.0 and get rid of the warning suppressions.

src/Featurizers/StandardScaleWrapperFeaturizer.h
```diff
@@ -0,0 +1,348 @@
...
+template <typename InputT, typename TransformedT>
+void StandardScalerTransformer<InputT, TransformedT>::execute_impl(typename BaseType::InputType const &input, typename BaseType::CallbackFunction const &callback) /*override*/ {
+    // ----------------------------------------------------------------------
+    using InputTraits                       = Traits<InputT>;
+    using TransformedTraits                 = Traits<TransformedT>;
+    // ----------------------------------------------------------------------
+
+    // if input is Nan for numeric types, it will be treated as missing value
+    // and return a Nan
+    if(InputTraits::IsNull(input)) {
```
davidbrownellWork: I don't think that we can make this assumption. There are scenarios where the caller might want to do this and other scenarios where null/NaN values are considered errors.

src/Featurizers/StandardScaleWrapperFeaturizer.h
```diff
@@ -0,0 +1,348 @@
...
+        callback(TransformedTraits::CreateNullValue());
+        return;
+    }
+
+#if (defined __clang__)
```
davidbrownellWork: I don't understand why these warnings need to be suppressed. I wouldn't expect to see any errors if the TransformedT is a double.

src/Featurizers/StandardScaleWrapperFeaturizer.h
```diff
@@ -0,0 +1,348 @@
...
+
+}
+
+// ----------------------------------------------------------------------
+// |
+// |  StandardScaleWrapperEstimator
+// |
+// ----------------------------------------------------------------------
+template <typename InputT, typename TransformedT, size_t MaxNumTrainingItemsV>
+StandardScaleWrapperEstimator<InputT, TransformedT, MaxNumTrainingItemsV>::StandardScaleWrapperEstimator(AnnotationMapsPtr pAllColumnAnnotations, size_t colIndex, bool with_mean, bool with_std) :
```
davidbrownellWork: Pick a naming convention and stick with it.

src/Featurizers/StandardScaleWrapperFeaturizer.h
```diff
@@ -0,0 +1,348 @@
...
+            Components::StandardDeviationEstimator<InputT, MaxNumTrainingItemsV>,
+            Details::StandardScaleWrapperEstimatorImpl<InputT, TransformedT, MaxNumTrainingItemsV>
+        >;
+
+    // ----------------------------------------------------------------------
+    // |
+    // |  Public Methods
+    // |
+    // ----------------------------------------------------------------------
+    StandardScaleWrapperEstimator(AnnotationMapsPtr pAllColumnAnnotations, size_t colIndex, bool with_mean, bool with_std);
```
davidbrownellWork: Use bit flags rather than bools, as it isn't clear at the call site what "true, true" or "false, true" actuall means.

src/Featurizers/UnitTests/StandardScaleWrapperFeaturizer_UnitTest.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+#include "../../3rdParty/optional.h"
+#include "../../Featurizers/StandardScaleWrapperFeaturizer.h"
+#include "../TestHelpers.h"
+#include "../../Traits.h"
+
+namespace NS = Microsoft::Featurizer;
+
+// test for no valid input
+TEST_CASE("no valid input") {
```
davidbrownellWork: I would expect to see each of these tests repeated for every flag permutation:

true, true
true,  false
false, true
false, false

src/Featurizers/Components/StandardDeviationEstimator.h
```diff
@@ -0,0 +1,261 @@
...
+            >,
+            MaxNumTrainingItemsV
+        >;
+
+    // ----------------------------------------------------------------------
+    // |
+    // |  Public Methods
+    // |
+    // ----------------------------------------------------------------------
+    StandardDeviationEstimator(AnnotationMapsPtr pAllColumnAnnotations, size_t colIndex);
```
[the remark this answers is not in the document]
angryjinyan: Implementation changed so that the user cannot specify average now.

src/Featurizers/Components/StandardDeviationEstimator.h
```diff
@@ -0,0 +1,261 @@
...
+        StandardStatisticalAnnotationData const &          data(StatisticalMetricsEstimator::get_annotation_data(estimator.get_column_annotations(), estimator.get_column_index(), StatisticalMetricsEstimatorName));
+
+        _average = data.Average;
+    }
+    return true;
+}
+ 
+template <typename T, typename StandardDeviationEstimatorT>
+void Details::StandardDeviationTrainingOnlyPolicy<T, StandardDeviationEstimatorT>::fit(InputType const &input) {
+    if(Microsoft::Featurizer::Traits<T>::IsNull(input))
```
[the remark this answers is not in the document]
angryjinyan: Add TODO comment. Strategies to deal with nulls will be implemented in the future

src/Featurizers/StandardScaleWrapperFeaturizer.h
angryjinyan: Combining IsNullableType and IsIntOrNumeric, only numerical types are eligible. Does writing like this harm understanding it?

src/Featurizers/StandardScaleWrapperFeaturizer.h
angryjinyan: TODO comment added. Will complete bit flag in a separate PR.

src/Featurizers/StandardScaleWrapperFeaturizer.h
angryjinyan: After discussion, adding IsNumeric will not be included in this PR.

src/Featurizers/StandardScaleWrapperFeaturizer.h
```diff
@@ -0,0 +1,331 @@
...
+        // ----------------------------------------------------------------------
+        using StandardStatisticalAnnotationData = Components::StandardStatisticalAnnotationData<InputT>;
+        using StatisticalMetricsEstimator       = Components::StatisticalMetricsEstimator<InputT, MaxNumTrainingItemsV>;
+
+        using StandardDeviationAnnotationData   = Components::StandardDeviationAnnotationData;
+        using StandardDeviationEstimator        = Components::StandardDeviationEstimator<InputT, MaxNumTrainingItemsV>;
+        // ----------------------------------------------------------------------
+        // if _withMean is false, average is set to 0
+        std::double_t average = 0;
+        // if _withStd is false, deviation is set to 1
```
michaelgsharp: Why do you set this to 1 if false, but set the average to 0 if its false?

src/Featurizers/StandardScaleWrapperFeaturizer.h
```diff
@@ -0,0 +1,331 @@
...
+            average = average_stats.Average;
+        }
+        if (_withStd) {
+            StandardDeviationAnnotationData   const &        deviation_stats(StandardDeviationEstimator::get_annotation_data(this->get_column_annotations(), _colIndex, Components::StandardDeviationEstimatorName));
+            deviation = static_cast<std::double_t>(deviation_stats.StandardDeviation);
+        }
+
+        // check if deviation is 0
+        // if it's 0, change it to 1
+        if (deviation == 0.0) {
```
michaelgsharp: why does the deviation need to be 1? 0 is a valid standard deviation.

src/Featurizers/StandardScaleWrapperFeaturizer.h
```diff
@@ -0,0 +1,331 @@
...
+    throw std::runtime_error("This will never be called");
+}
+
+template <typename InputT, typename TransformedT, size_t MaxNumTrainingItemsV>
+void Details::StandardScaleWrapperEstimatorImpl<InputT, TransformedT, MaxNumTrainingItemsV>::complete_training_impl(void) /*override*/ {
+}
+
+}
+}
+}
```
michaelgsharp: Can you add namespace comments here please.

src/Featurizers/StandardScaleWrapperFeaturizer.h
angryjinyan: Because if withStd and withMean are set to false, default operation is to scale data centered at 0 with unit standard deviation. So average is set to 0 and standard deviation is set to 1.

src/Featurizers/StandardScaleWrapperFeaturizer.h
angryjinyan: 0 is a valid standard deviation indeed but in this featurizer, standard deviation need to be devided to get unit standard deviation and 0 is not a valid denominator.

src/Featurizers/StandardScaleWrapperFeaturizer.h
angryjinyan: namespace comments are added.

PR #40, opened 2019-12-17
Fixes for ORT

michaelgsharp:
Fixed Nullable Type issue for ImputationMarkerFeaturizer and MissingDummiesFeaturizer by making transformers take normal types. Same fix as was used with CatImputer
Fixed some ORT code gen issues by removing `_t` and making sure its `std::string`

### DIFF
--- a/src/Featurizers/ImputationMarkerFeaturizer.h
+++ b/src/Featurizers/ImputationMarkerFeaturizer.h
@@ -12,19 +12,30 @@ namespace Featurizer {
 namespace Featurizers {
 
 /////////////////////////////////////////////////////////////////////////
+///  \class        ImputationMarkerTraits
+///  \brief         Traits for mapping the input/output types for the ImputationMarker.
+///                 This allows us to only change one place if typings need to change.
+///
+template <typename TransformedT>
+struct ImputationMarkerTraits{
+    using InputType = typename Traits<TransformedT>::nullable_type;
+
+    static_assert(Traits<InputType>::IsNullableType, "'InputT' must be a nullable type");
+};
+
+/////////////////////////////////////////////////////////////////////////
 ///  \class         ImputationMarkerTransformer
 ///  \brief         if input is Null, return true. Otherwise return false
 ///
-template <typename T>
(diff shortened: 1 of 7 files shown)


src/Featurizers/ImputationMarkerFeaturizer.h
```diff
@@ -11,20 +11,31 @@ namespace Microsoft {
 namespace Featurizer {
 namespace Featurizers {
 
+/////////////////////////////////////////////////////////////////////////
+///  \class        ImputationMarkerTraits
+///  \brief         Traits for mapping the input/output types for the ImputationMarker.
+///                 This allows us to only change one place if typings need to change.
+///
+template <typename TransformedT>
+struct ImputationMarkerTraits{
```
yuslepukhin: >ImputationMarkerTraits [](start = 7, length = 22)

Should this be generalized since the same code is used at least in two places?

michaelgsharp: I am not sure if the input/output type mappings will be the same for each featurizer which is why I have them separate for each one. Currently that is the case, just not sure about always. @davidbrownellWork do you know?

michaelgsharp: I moved this code into a general location and have update the transformers to they use that one.

src/Traits.h
```diff
@@ -1158,6 +1158,20 @@ struct Traits<nonstd::optional<T>>  {
...
+/////////////////////////////////////////////////////////////////////////
+///  \class        MakeNullabeType
+///  \brief         We have many several situations where the transformer operates on nullable types,
+///                 but we don't want to instantiate the class as Transformer<optional<type>>.
+///                 This struct allows us to instantiate the class as Transformer<type>, but still be able to
+///                 operate on optional<type> and to make sure that optional<type> is not initially passed in.
+///
+template <typename T>
+struct MakeNullableType{
+    using InputType = typename Traits<T>::nullable_type;
```
davidbrownellWork: This class can be used for both InputTypes and TransformedTypes; it is probably better to change the name from `InputType` to `type` to remain consistent with other C++ type decorators.

PR #50, opened 2020-01-02
Updates to the featurizer YAML to make types more generic (less specific to C++)

davidbrownellWork:
This change is in preparation for the introduction of the datetime type and the corresponding code generator plugins.

- ONNX Runtime types now use "_t" types where the macro generates C++ code (this wasn't applied consistently before this change)

### DIFF
--- a/src/SharedLibrary/GeneratedCode/SharedLibrary_CatImputerFeaturizer.cpp
+++ b/src/SharedLibrary/GeneratedCode/SharedLibrary_CatImputerFeaturizer.cpp
@@ -24,10 +24,10 @@ extern "C" {
 
 /* ---------------------------------------------------------------------- */
 /* |                                                                      */
-/* |  CatImputerFeaturizer <int8_t> */
+/* |  CatImputerFeaturizer <int8> */
 /* |                                                                      */
 /* ---------------------------------------------------------------------- */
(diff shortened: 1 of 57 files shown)


yuslepukhin:
>       float execute(float const &value);

I think primitive types are better passed by value. A single push is always cheaper than LEA;PUSH. The other parts I am not sure  about it. It is important that C++ context is not broken by the removal of _t.

---
Refers to: src/Tools/CodeGenerator/Featurizers.yaml:22 in 12e0361. [](commit_id = 12e03619d3faa29adfe3bf411b7fe6518c0fadb2, deletion_comment = False)

davidbrownellWork:
> 
> 
> > ```
> >   float execute(float const &value);
> > ```
> 
> I think primitive types are better passed by value. A single push is always cheaper than LEA;PUSH. The other parts I am not sure about it. It is important that C++ context is not broken by the removal of _t.
> 
> Refers to: src/Tools/CodeGenerator/Featurizers.yaml:22 in [12e0361](https://github.com/microsoft/FeaturizersLibrary/commit/12e03619d3faa29adfe3bf411b7fe6518c0fadb2). [](commit_id = [12e0361](https://github.com/microsoft/FeaturizersLibrary/commit/12e03619d3faa29adfe3bf411b7fe6518c0fadb2), deletion_comment = False)

Let's address this as part of a separate checkin. Given that this code is dependent upon the input type, the way I am thinking about this involves something like https://www.boost.org/doc/libs/1_66_0/libs/utility/call_traits.htm, but there may be a better way. Let's discuss and determine a path forward.

PR #52, opened 2020-01-02
update Featurizers.yaml

wangyems:


wangyems:
Based on offline chat, will update this PR after featurizer is in

PR #54, opened 2020-01-06
DateTimeFeaturizer now operates on std::chrono::system_clock::time_point

davidbrownellWork:


src/Featurizers/DateTimeFeaturizer.cpp
```diff
@@ -371,19 +379,19 @@ void DateTimeTransformer::save(Archive & ar) const /*override*/ {
...
-        InputType dateKey = (input - input % 86400);
-
-        auto x = _dateHolidayMap.find(dateKey);
-        if (x != _dateHolidayMap.end()) {
-            result.holidayName = x->second;
-        }
+        // Values in the holiday map are based on midnight of the corresponding day. When looking up holidays for this time,
+        // remove any seconds since midnight.
+        long long const                     totalSeconds(std::chrono::time_point_cast<std::chrono::seconds>(input).time_since_epoch().count());
+        long long const                     holidayKey(totalSeconds - (totalSeconds % secondsPerDay));
```
yuslepukhin: >holidayKey(totalSeconds - (totalSeconds % secondsPerDay)); [](start = 44, length = 58)

Would (totalSeconds/secondsPerDay) * secondsPerDay be a better arithmetics?

src/Featurizers/DateTimeFeaturizer.cpp
```diff
@@ -371,19 +379,19 @@ void DateTimeTransformer::save(Archive & ar) const /*override*/ {
...
+        HolidayMap::const_iterator const    iter(_dateHolidayMap.find(static_cast<std::int64_t>(holidayKey)));
```
yuslepukhin: >static_cast<std::int64_t> [](start = 70, length = 25)

I'd move that to totalSeconds and convert everything to int64_t?

yuslepukhin:

Not seeing anything here that would affect ORT tests.

---
Refers to: src/Featurizers/UnitTests/DateTimeFeaturizer_UnitTests.cpp:4 in 73ff101. [](commit_id = 73ff101219fdc206083c3c657a64da4508ac7eb2, deletion_comment = False)

src/Featurizers/DateTimeFeaturizer.cpp
davidbrownellWork: I'm not sure I completely follow your comment, but I think that you are saying that either the map key should be long long or the holidaySeconds should be int64?

src/SharedLibrary/GeneratedCode/SharedLibrary_Common.h
```diff
@@ -50,6 +50,36 @@ FEATURIZER_LIBRARY_API bool DestroyErrorInfoString(/*in*/ char const *input_ptr,
 FEATURIZER_LIBRARY_API bool DestroyErrorInfo(/*in*/ ErrorInfoHandle *pHandle);
 FEATURIZER_LIBRARY_API bool DestroyTransformerSaveData(/*in*/ unsigned char const *pBuffer, /*in*/ std::size_t cBufferSize, /*out*/ ErrorInfoHandle **ppErrorInfo);
 
+FEATURIZER_LIBRARY_API_PACK_PREFIX;
+
+struct DateTimeParameter {
```
michaelgsharp: The documentation we looked at was a plain union. This is a union within struct. I think it should still work, but I wanted to mention it just in case.

src/SharedLibrary/GeneratedCode/SharedLibrary_HashOneHotVectorizerFeaturizer.cpp
```diff
@@ -12,6 +12,7 @@
 
 // These method(s) are defined in SharedLibrary_Common.cpp
 ErrorInfoHandle * CreateErrorInfo(std::exception const &ex);
+std::chrono::system_clock::time_point CreateDateTime(DateTimeParameter const &param);
```
michaelgsharp: Is this going to be in all the `SharedLibrary` code even if they don't use DateTIme stuff?

src/Tools/CodeGenerator/Featurizers.yaml
```diff
@@ -104,7 +104,7 @@
       }
 
   type_mappings:
-    - input_type: int64
+    - input_type: datetime
```
michaelgsharp: will this field always represent the input for the shared library as opposed to the input for the featurizer itself?

src/SharedLibrary/GeneratedCode/SharedLibrary_Common.h
```diff
@@ -50,6 +50,36 @@ FEATURIZER_LIBRARY_API bool DestroyErrorInfoString(/*in*/ char const *input_ptr,
 FEATURIZER_LIBRARY_API bool DestroyErrorInfo(/*in*/ ErrorInfoHandle *pHandle);
 FEATURIZER_LIBRARY_API bool DestroyTransformerSaveData(/*in*/ unsigned char const *pBuffer, /*in*/ std::size_t cBufferSize, /*out*/ ErrorInfoHandle **ppErrorInfo);
 
+FEATURIZER_LIBRARY_API_PACK_PREFIX;
+
+struct DateTimeParameter {
+    // ----------------------------------------------------------------------
+    // |  Public Types
+    enum DateTimeTypeValue {
+        DateTimeInt64,      // Posix time
```
michaelgsharp: Can we add explicit numbers to these starting with 1 so it matches the other enums we have?

src/SharedLibrary/GeneratedCode/SharedLibrary_Common.h
davidbrownellWork: The binary layout should still be the same, so I think that we will be all right. This is something that we should keep in mind if we start to see test failures in ML.Net.

src/Tools/CodeGenerator/Featurizers.yaml
davidbrownellWork: The meaning of datetime changes based on context...

- From the shared library's perspective / C++, it is: std::chrono::system_clock::time_point
- From the C interface's perspective, it is: ISO 8601 string, posix time (int64)
- From the ML.Net interface's perspective, it is: System.DateTime, ISO 8601 string, and (potentially) posix time (int64)
- From a potential Python interface's perspective, it is: datetime.datetime, ISO 8601 string, and (potentially) posix time (int64)

src/SharedLibrary/GeneratedCode/SharedLibrary_HashOneHotVectorizerFeaturizer.cpp
davidbrownellWork: Yes, however the linker will remove the code if it isn't being used.

PR #58, opened 2020-01-10
add submodule checkout to build pipeline

wangyems:


CI/templates/BuildAndTest.job_template.yaml
```diff
@@ -25,6 +25,9 @@ jobs:
       job_display_name: "${{ parameters.job_display_name }}"
       operating_system: ${{ parameters.operating_system }}
       steps:
+        - checkout: self
```
davidbrownellWork: We will need these same changes for every collection of steps that doesn't explicitly disable the checkout (where checkout isn't set to None). Can you analyze all of the build files to ensure that they are consistent?

As an alternative to this, we can update _job_template.yaml to take a parameter that controls a checkout step added to the steps - if it is set to false, checkout is set to None. Otherwise checkout has these updates.

Let's chat in person if you want to discuss potential designs for this.

CI/templates/_job_template.yaml
```diff
@@ -9,6 +9,10 @@ parameters:
   job_display_name: ""
   operating_system: "" # Windows|Linux|MacOS
   steps: []
+  submodule_checkout:
```
davidbrownellWork: There are pipelines that use this template (see the Publish pipelines) that won't want to checkout the code. Those yaml files should be updated to override this parameter.

PR #64, opened 2020-01-13
TruncatedSVDFeaturizer

wangyems:


src/Featurizers/TruncatedSVDFeaturizer.h
```diff
@@ -0,0 +1,631 @@
...
+#   pragma clang diagnostic ignored "-Wdouble-promotion"
+#   pragma clang diagnostic ignored "-Wcast-align"
+#   pragma clang diagnostic ignored "-Wfloat-equal"
+#   pragma clang diagnostic ignored "-Wshadow"
+#elif (defined _MSC_VER)
+#   pragma warning(push)
+#   pragma warning(disable: 4127)
+#endif
+
+#include "../3rdParty/eigen/Eigen/Dense"
```
yuslepukhin: >../3rdParty/eigen/Eigen/Dense" [](start = 10, length = 30)

Would you rather just state #include "Eigen/Dense" ? For this 1) In the main CMakeLists.txt add interface only target:
    add_library(eigen INTERFACE)
    target_include_directories(eigen INTERFACE ${CMAKE_CURRENT_SOURCE_DIR}/3rdParty/eigen)

src/Featurizers/TruncatedSVDFeaturizer.h
```diff
@@ -0,0 +1,631 @@
...
+#   pragma clang diagnostic pop
+#elif (defined _MSC_VER)
+#   pragma warning(pop)
+#endif
+
+namespace Microsoft {
+namespace Featurizer {
+namespace Featurizers {
+
+static constexpr char const * const         SVDComponentsEstimatorName("SVDComponentsEstimator");
```
yuslepukhin: >static [](start = 0, length = 6)

static means local to the current CPP file. When including this you are creating multiple definitions of the same string in multiple CPP files that are not related to your implementation. Please, put this into a CPP file and declare here an extern.

src/Featurizers/TruncatedSVDFeaturizer.h
```diff
@@ -0,0 +1,631 @@
...
+ *    documentation and/or other materials provided with the distribution.
+ * 
+ * 3. Neither the name of the authors nor the names of its contributors
+ *    may be used to endorse or promote products derived from this
+ *    software without specific prior written permission.
+ */
+template<typename Scalar>
+inline void sample_gaussian(Scalar& x, Scalar& y)
+{
+	using std::sqrt;
```
yuslepukhin: >	 [](start = 0, length = 1)

spaces

src/Featurizers/TruncatedSVDFeaturizer.h
```diff
@@ -0,0 +1,631 @@
...
+    // ----------------------------------------------------------------------
+    SVDComponentsAnnotationData(T singularvalues, T singularvectors);
+    ~SVDComponentsAnnotationData(void) = default;
+
+    FEATURIZER_MOVE_CONSTRUCTOR_ONLY(SVDComponentsAnnotationData);
+};
+
+namespace ComponentsDetails {
+
+namespace {
```
yuslepukhin: This also has a meaning of static, unique to the current unit of compilation. You have already introduced ComponentDetails namespace for the purpose. What is the reason for nameless namespaces? What will it do in difference CC files where it is included?

src/Featurizers/TruncatedSVDFeaturizer.h
```diff
@@ -0,0 +1,631 @@
...
+SVDComponentsAnnotationData<TransformedT> ComponentsDetails::SVDTrainingOnlyPolicy<InputT, TransformedT>::complete_training(void) {
+
+    //the following code in this function is introduced from RedSVD
+    typedef typename TransformedT::Scalar Scalar;
+    typedef typename Eigen::Matrix<Scalar, Eigen::Dynamic, Eigen::Dynamic> DenseMatrix;
+
+    Eigen::Index rank = (_matrix.rows() < _matrix.cols()) ? _matrix.rows() : _matrix.cols();
+
+    // Gaussian Random Matrix for _matrix^T
+	DenseMatrix O(_matrix.rows(), rank);
```
yuslepukhin: >	 [](start = 0, length = 1)

spaces instead of tabs

src/Featurizers/TruncatedSVDFeaturizer.h
```diff
@@ -0,0 +1,631 @@
...
+	sample_gaussian(O);
+			
+	// Compute Sample Matrix of _matrix^T
+	DenseMatrix Y = _matrix.transpose() * std::move(O);
```
yuslepukhin: >std::move(O) [](start = 39, length = 12)

What's the purpose of the move? Same below.

src/Featurizers/TruncatedSVDFeaturizer.h
```diff
@@ -0,0 +1,631 @@
...
+    return true;
+}
+
+template <typename InputT, typename TransformedT>
+void SVDTransformer<InputT, TransformedT>::save(Archive &ar) const /*override*/ {
+    // Version
+    Traits<std::uint16_t>::serialize(ar, 1); // Major
+    Traits<std::uint16_t>::serialize(ar, 0); // Minor
+
+    // Data TODO
```
yuslepukhin: >/ Data TODO [](start = 5, length = 11)

Will transformer function w/o training data?

src/Featurizers/TruncatedSVDFeaturizer.h
```diff
@@ -0,0 +1,631 @@
...
+// |
+// ----------------------------------------------------------------------
+template <typename InputT, typename TransformedT>
+SVDTransformer<InputT, TransformedT>::SVDTransformer(TransformedT singularvalues, TransformedT singularvectors) :
+    //no validation here because same variable has been validated in annotation class
+    _singularvalues(std::move(singularvalues)),
+    _singularvectors(std::move(singularvectors)) {
+}
+
+//TODO:
```
yuslepukhin: Can we remove commented code?

src/Featurizers/UnitTests/TruncatedSVDFeaturizer_UnitTests.cpp
```diff
@@ -0,0 +1,192 @@
...
+    CHECK((output - label).norm() < eps);
+   
+#if (defined __clang__)
+#   pragma clang diagnostic pop
+#endif
+}
+
+//TODO
+TEST_CASE("Serialization/Deserialization") {
+
```
yuslepukhin: Seems incomplete.

src/Featurizers/TruncatedSVDFeaturizer.h
wangyems: Just save some spaces as the variable will no longer needed

src/Featurizers/TruncatedSVDFeaturizer.h
wangyems: transformer function will contain the components obtained from training data(singularvalues and singularvectors in this featurizer). But I don't think ORT need TODO part(serialization/deserialization) @davidbrownellWork is that correct?

yuslepukhin: All Transformers in ORT are instantiated using deserialization.

---
In reply to: [366112663](https://github.com/microsoft/FeaturizersLibrary/pull/64#discussion_r366112663) [](ancestors = 366112663)

src/Featurizers/TruncatedSVDFeaturizer.h
yuslepukhin: This part I do not understand.

---
In reply to: [366111577](https://github.com/microsoft/FeaturizersLibrary/pull/64#discussion_r366111577) [](ancestors = 366111577)

src/Featurizers/UnitTests/CMakeLists.txt
```diff
@@ -52,6 +53,7 @@ foreach(_test_name IN ITEMS
 
     target_link_libraries(${_test_name} PRIVATE
         FeaturizersCode
+        eigen
```
yuslepukhin: > eigen [](start = 7, length = 6)

You don't need to link to INTERFACE libraries.

src/Featurizers/UnitTests/TruncatedSVDFeaturizer_UnitTests.cpp
```diff
@@ -0,0 +1,192 @@
...
+                   TransformedT const &singularVectorsLabel,
+                   std::float_t eps = 0.0001f) {
+
+    using SVDComponentsEstimator                       = NS::Featurizers::SVDComponentsEstimator<InputT, TransformedT>;
+    using SVDComponentsAnnotationData                  = NS::Featurizers::SVDComponentsAnnotationData<TransformedT>;
+
+    NS::AnnotationMapsPtr const                        pAllColumnAnnotations(NS::CreateTestAnnotationMapsPtr(1));
+    SVDComponentsEstimator                             estimator(pAllColumnAnnotations, 0);
+
+    estimator.begin_training();
```
yuslepukhin: >estimator.begin_training(); [](start = 4, length = 27)

Is it possible to use TestHelper method or its specialization Train<>()?

src/Featurizers/TruncatedSVDFeaturizer.h
```diff
@@ -0,0 +1,631 @@
...
+#   pragma clang diagnostic push
+#   pragma clang diagnostic ignored "-Wdouble-promotion"
+#endif
+		
+    const Scalar PI(3.1415926535897932384626433832795028841971693993751f);
+
+#if (defined __clang__)
+#   pragma clang diagnostic pop
+#endif
+
```
yuslepukhin: Tabs

src/Featurizers/TruncatedSVDFeaturizer.h
wangyems: Will add serialization/deserialization in Traits in another PR

src/Featurizers/UnitTests/TruncatedSVDFeaturizer_UnitTests.cpp
wangyems: For now it's not, but I can extend that function

src/Featurizers/TruncatedSVDFeaturizer.h
davidbrownellWork: This is a constexpr value, which requires immediate assignment. Perhaps the "static" can be removed?

src/Featurizers/TruncatedSVDFeaturizer.h
davidbrownellWork: We should introduce partially created Featurizers into master; if you implemented matrix serialization in a separate PR, make sure that one goes in before this one.

src/Featurizers/TruncatedSVDFeaturizer.h
wangyems: sure

src/Featurizers/UnitTests/CMakeLists.txt
wangyems: looks the compiler is not happy without that(can't find Eigen/Dense)

src/Featurizers/TruncatedSVDFeaturizer.h
```diff
@@ -0,0 +1,631 @@
...
+    FEATURIZER_MOVE_CONSTRUCTOR_ONLY(SVDComponentsAnnotationData);
+};
+
+namespace ComponentsDetails {
+
+/////////////////////////////////////////////////////////////////////////
+///  \class         SVDTrainingOnlyPolicy
+///  \brief         `SVDComponentsEstimator` implementation details.
+///
+template <typename InputT, typename TransformedT>
```
davidbrownellWork: If InputT and TransformedT are the same, let's make them a single template type.

src/Featurizers/TruncatedSVDFeaturizer.h
```diff
@@ -0,0 +1,631 @@
...
+        ComponentsDetails::SVDTrainingOnlyPolicy<InputT, TransformedT>,
+        MaxNumTrainingItemsV
+    > {
+public:
+    // ----------------------------------------------------------------------
+    // |
+    // |  Public Types
+    // |
+    // ----------------------------------------------------------------------
+    using BaseType =
```
davidbrownellWork: If there are restrictions on the types, assert that here. If it is possible to derive the TransformedT, do that in the code rather than requesting that information from the user.

Based on our in-person conversation, it sounds like the input is restricted to numeric and integer types. Think also about calculating the TransformedT based on the InputT, thereby eliminating the need for that type.

src/Featurizers/TruncatedSVDFeaturizer.h
```diff
@@ -0,0 +1,631 @@
...
+///  \class         SVDComponentsEstimator
+///  \brief         This class generates SVDComponentsAnnotationData
+///                 by processing input matrix using Eigen
+///
+template <
+    typename InputT,
+    typename TransformedT,
+    size_t MaxNumTrainingItemsV=std::numeric_limits<size_t>::max()
+>
+class SVDComponentsEstimator :
```
davidbrownellWork: Could the functionality in this Estimator be implemented in the Impl class? If so, you can remove this estimator and eliminate the use of the PipelineExecutionEstimatorImpl.

src/Featurizers/TruncatedSVDFeaturizer.h
```diff
@@ -0,0 +1,631 @@
...
+///  \class         SVDEstimatorImpl
+///  \brief         Estimator that reads an annotation created by the `SVDComponentsEstimator`
+///                 and creates a `SVDTransformer` object.
+///
+template <
+    typename InputT,
+    typename TransformedT,
+    size_t MaxNumTrainingItemsV=std::numeric_limits<size_t>::max()
+>
+////InputT and TransformedT are Containers type
```
davidbrownellWork: Does this comment still apply?

src/Featurizers/TruncatedSVDFeaturizer.h
```diff
@@ -0,0 +1,631 @@
...
+        SVDComponentsAnnotationData const &              data(SVDComponentsEstimator::get_annotation_data(this->get_column_annotations(), _colIndex, SVDComponentsEstimatorName));
+
+        return typename BaseType::TransformerUniquePtr(new SVDTransformer<InputT, TransformedT>(data.SingularValues, data.SingularVectors));
+    }
+};
+
+} // namespace Details
+
+/////////////////////////////////////////////////////////////////////////
+///  \class         SVDFeaturizer
```
davidbrownellWork: This class name doesn't match with the defined class.

src/Featurizers/TruncatedSVDFeaturizer.h
```diff
@@ -0,0 +1,631 @@
...
+// |
+// |  SVDComponentsAnnotationData
+// |
+// ----------------------------------------------------------------------
+template <typename T>
+SVDComponentsAnnotationData<T>::SVDComponentsAnnotationData(T singularvalues, T singularvectors) :
+    SingularValues(
+        std::move(
+            [&singularvalues](void) -> T & {
+                if(singularvalues.size() == 0)
```
davidbrownellWork: singularvalues.empty() is probably a more clear description of the check being done here.

src/Featurizers/TruncatedSVDFeaturizer.h
```diff
@@ -0,0 +1,631 @@
...
+// ----------------------------------------------------------------------
+template <typename InputT, typename TransformedT>
+void SVDTransformer<InputT, TransformedT>::execute_impl(typename BaseType::InputType const &input, typename BaseType::CallbackFunction const &callback) /*override*/ {
+    if (input.rows() == 0)
+        throw std::invalid_argument("Input matrix rows() invalid");
+    
+    if (input.cols() != _singularvectors.rows()) 
+        throw std::invalid_argument("Input matrix cols() invalid");
+
+    callback(std::move(input * _singularvectors));
```
davidbrownellWork: `input * _singlevectors` will create a temporary (which isn't assigned to a variable), so `move` isn't necessary here.

src/Featurizers/UnitTests/TruncatedSVDFeaturizer_UnitTests.cpp
davidbrownellWork: I agree with Dmitri - we should use or extend Train rather than inventing a one-off solution.

yuslepukhin:
> add_library(FeaturizersCode STATIC

Add your source header to the list of the files so it shows up in IDE

---
Refers to: src/Featurizers/cmake/FeaturizersCode.cmake:7 in 43c08bc. [](commit_id = 43c08bcad6db0e2b59972db2b63e9fc1e213bfa0, deletion_comment = False)

src/Featurizers/TruncatedSVDFeaturizer.h
```diff
@@ -0,0 +1,631 @@
...
+//             TransformedT                   singularvectors(Traits<TransformedT>::deserialize(ar));
+
+//             return SVDTransformer<InputT, TransformedT>(std::move(singularvalues), std::move(singularvectors));
+//         }()
+//     ) {
+// }
+
+template <typename InputT, typename TransformedT>
+bool SVDTransformer<InputT, TransformedT>::operator==(SVDTransformer const &other) const {
+    if ((this->_singularvalues - other._singularvectors).norm() > 0.000001f || (this->_singularvalues - other._singularvectors).norm() > 0.000001f)
```
yuslepukhin: >if ((this->_singularvalues - other._singularvectors).norm() > 0.000001f || (this->_singularvalues - other._singularvectors).norm() > 0.000001f) [](start = 4, length = 143)

I think something is off here.

wangyems: It's a bug. Thanks for pointing that out. Already been corrected and will be in the next update

src/Featurizers/TruncatedSVDFeaturizer.h
wangyems: not any more, will remove

src/Featurizers/TruncatedSVDFeaturizer.h
wangyems: Eigen::Matrix does not have a empty() method

src/UnitTests/Traits_UnitTest.cpp
```diff
@@ -409,7 +409,6 @@ TEST_CASE("Serialization") {
     matrix(0, 0) = 1.0f;
     matrix(0, 1) = 0.0f;
     CHECK(SerializationTestImpl(Eigen::MatrixX<float>()));
-    CHECK(SerializationTestImpl(Eigen::MatrixX<float>(1, 2)));
```
davidbrownellWork: It sounds like this change is to address a build issue. If so, it should be a part of its own PR. Lets attempt to root cause the issue before making changes to remove tests.

src/Featurizers/TruncatedSVDFeaturizer.h
```diff
@@ -0,0 +1,382 @@
...
+ * 2. Redistributions in binary form must reproduce the above Copyright
+ *    notice, this list of conditions and the following disclaimer in the
+ *    documentation and/or other materials provided with the distribution.
+ *
+ * 3. Neither the name of the authors nor the names of its contributors
+ *    may be used to endorse or promote products derived from this
+ *    software without specific prior written permission.
+ */
+template<typename Scalar>
+inline void sample_gaussian(Scalar& x, Scalar& y)
```
davidbrownellWork: >sample_gaussian [](start = 12, length = 15)

It looks like these functions are only used in the implementation. If that is the case, they can be moved until after the declaration of the different classes but before their declarations.

src/Featurizers/TruncatedSVDFeaturizer.h
```diff
@@ -0,0 +1,382 @@
...
+    gram_schmidt(Z);
+
+    // Range(C) = Range(B)
+    DenseMatrix C = Z.transpose() * B;
+
+    Eigen::JacobiSVD<DenseMatrix> svdOfC(C, Eigen::ComputeThinV);
+
+    // C = USV^T
+    // A = Z * U * S * V^T * Y^T()
+    this->_singularValues = svdOfC.singularValues();
```
davidbrownellWork: It looks like _matrix isn't used anymore. We should free any memory associated with that object.

src/Featurizers/TruncatedSVDFeaturizer.h
```diff
@@ -0,0 +1,382 @@
...
+    FitResult fit_impl(typename BaseType::InputType const *pBuffer, size_t) override {
+        _matrix = *pBuffer;
+
+        return FitResult::Continue;
+    }
+    void complete_training_impl(void) override;
+
+    // MSVC has problems when the definition is separate from the declaration
+    typename BaseType::TransformerUniquePtr create_transformer_impl(void) override {
+        return typename BaseType::TransformerUniquePtr(new TruncatedSVDTransformer<MatrixT>(_singularValues, _singularVectors));
```
davidbrownellWork: >_singularValues, _singularVectors [](start = 92, length = 33)

Should these values be moved into TrancatedSVDTransformer?

src/Featurizers/TruncatedSVDFeaturizer.h
```diff
@@ -0,0 +1,382 @@
...
+
+    // ----------------------------------------------------------------------
+    // |
+    // |  Private Methods
+    // |
+    // ----------------------------------------------------------------------
+    bool begin_training_impl(void) override;
+    // MSVC has problems when the declaration and definition are separated
+    FitResult fit_impl(typename BaseType::InputType const *pBuffer, size_t) override {
+        _matrix = *pBuffer;
```
davidbrownellWork: >_matrix = *pBuffer; [](start = 8, length = 19)

This line concerns me. If training really only supports 1 input, then we should have code that throws an exception if multiple are found. However, let's deep dive on this to ensure that the desired behavior is correctly implemented.

src/Featurizers/TruncatedSVDFeaturizer.h
```diff
@@ -0,0 +1,392 @@
+// ----------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License
+// ----------------------------------------------------------------------
+#pragma once
+
+#include "../Featurizer.h"
```
yuslepukhin: >include "../Featurizer.h" [](start = 1, length = 25)

This file should be added to CMake file

src/Featurizers/TruncatedSVDFeaturizer.h
```diff
@@ -0,0 +1,392 @@
...
+    // ----------------------------------------------------------------------
+    using BaseType                                   = StandardTransformer<MatrixT, MatrixT>;
+
+    // ----------------------------------------------------------------------
+    // |
+    // |  Public Methods
+    // |
+    // ----------------------------------------------------------------------
+    TruncatedSVDTransformer(MatrixT singularvalues, MatrixT singularvectors);
+    TruncatedSVDTransformer(Archive &ar);
```
yuslepukhin: > Tr [](start = 3, length = 3)

explicit

src/Featurizers/TruncatedSVDFeaturizer.h
```diff
@@ -0,0 +1,392 @@
...
+// ----------------------------------------------------------------------
+// |
+// |  TruncatedSVDTransformer
+// |
+// ----------------------------------------------------------------------
+template <typename MatrixT>
+TruncatedSVDTransformer<MatrixT>::TruncatedSVDTransformer(MatrixT singularvalues, MatrixT singularvectors) :
+    _singularvalues(
+        std::move(
+            [&singularvalues](void) -> MatrixT & {
```
yuslepukhin: > [&singularvalues] [](start = 11, length = 18)

These unnecessary lambdas hurt my eye

src/Featurizers/TruncatedSVDFeaturizer.h
wangyems: Will be fixed in a future refactor

PR #65, opened 2020-01-14
StandardScaleWrapper shared library implementation

angryjinyan:
Finish SSW shared library implementation

### DIFF
--- a/src/SharedLibrary/IntegrationTests/SharedLibrary_SSWFeaturizer_IntegrationTest.cpp
+++ b/src/SharedLibrary/IntegrationTests/SharedLibrary_SSWFeaturizer_IntegrationTest.cpp
@@ -10,75 +10,75 @@
 TEST_CASE("int8_t_with_mean_with_std") {
     StandardScaleWrapperFeaturizer_int8_Test(
         std::vector<std::int8_t>{
-            static_cast<std::int8_t>(0),
-            static_cast<std::int8_t>(0),
-            static_cast<std::int8_t>(1),
-            static_cast<std::int8_t>(1)
+            0,
+            0,
+            1,
+            1
         },
         std::vector<std::int8_t>{
-            static_cast<std::int8_t>(2)
+            2
         },
         [](std::vector<std::double_t> const &args) {
-            return args == std::vector<std::double_t>{static_cast<std::double_t>(3)};
+            return args[0] == 3.0;
         },
-        bool(true),
-        bool(true)
+        true,
+        true
     );
 }
 
 TEST_CASE("double_t_without_mean_with_std") {
     StandardScaleWrapperFeaturizer_double_Test(
         std::vector<std::double_t>{
-            static_cast<std::double_t>(0),
-            static_cast<std::double_t>(0),
-            static_cast<std::double_t>(1),
-            static_cast<std::double_t>(1)
+            0.0,
+            0.0,
+            1.0,
+            1.0
         },
         std::vector<std::double_t>{
-            static_cast<std::double_t>(2)
+            2.0
         },
         [](std::vector<std::double_t> const &args) {
-            return args == std::vector<std::double_t>{static_cast<std::double_t>(4)};
+            return args[0] == 4.0;
         },
-        bool(false),
-        bool(true)
+        false,
+        true
     );
 }
 
 TEST_CASE("double_t_with_mean_without_std") {
     StandardScaleWrapperFeaturizer_double_Test(
         std::vector<std::double_t>{
-            static_cast<std::double_t>(0),
-            static_cast<std::double_t>(0),
-            static_cast<std::double_t>(1),
-            static_cast<std::double_t>(1)
+            0.0,
+            0.0,
+            1.0,
+            1.0
         },
         std::vector<std::double_t>{
-            static_cast<std::double_t>(2)
+            2.0
         },
         [](std::vector<std::double_t> const &args) {
-            return args == std::vector<std::double_t>{static_cast<std::double_t>(1.5)};
+            return args[0] == 1.5;
         },
-        bool(true),
-        bool(false)
+        true,
+        false
     );
 }
 
 TEST_CASE("double_t_without_mean_without_std") {
     StandardScaleWrapperFeaturizer_double_Test(
         std::vector<std::double_t>{
-            static_cast<std::double_t>(0),
-            static_cast<std::double_t>(0),
-            static_cast<std::double_t>(1),
-            static_cast<std::double_t>(1)
+            0.0,
+            0.0,
+            1.0,
+            1.0
         },
         std::vector<std::double_t>{
-            static_cast<std::double_t>(2)
+            2.0
         },
         [](std::vector<std::double_t> const &args) {
-            return args == std::vector<std::double_t>{static_cast<std::double_t>(2)};
+            return args[0] == 2.0;
         },
-        bool(false),
-        bool(false)
+        false,
+        false
     );
 }
--- a/src/SharedLibrary/IntegrationTests/SharedLibrary_SSWFeaturizer_IntegrationTest.cpp
+++ b/src/SharedLibrary/IntegrationTests/SharedLibrary_SSWFeaturizer_IntegrationTest.cpp
@@ -1,0 +1,84 @@
+// ----------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License
+// ----------------------------------------------------------------------
+#define CATCH_CONFIG_MAIN
+#include "catch.hpp"
+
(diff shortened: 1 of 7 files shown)


src/SharedLibrary/IntegrationTests/SharedLibrary_SSWFeaturizer_IntegrationTest.cpp
```diff
@@ -0,0 +1,84 @@
...
+            static_cast<std::int8_t>(1),
+            static_cast<std::int8_t>(1)
+        },
+        std::vector<std::int8_t>{
+            static_cast<std::int8_t>(2)
+        },
+        [](std::vector<std::double_t> const &args) {
+            return args == std::vector<std::double_t>{static_cast<std::double_t>(3)};
+        },
+        bool(true),
```
yuslepukhin: > bool(true) [](start = 7, length = 11)

What does this mean?

yuslepukhin:
>         return true;

Suggest moving this to the end, then there would be no unreachable code.

---
Refers to: src/SharedLibrary/GeneratedCode/SharedLibrary_StandardScaleWrapperFeaturizer.cpp:48 in 85e8cce. [](commit_id = 85e8ccecccfbbcba1034cf1794a176fc80d8d973, deletion_comment = False)

yuslepukhin:
>         return false;

Is this a normal operation or there should be an assert?

---
Refers to: src/SharedLibrary/GeneratedCode/SharedLibrary_StandardScaleWrapperFeaturizer.cpp:33 in 85e8cce. [](commit_id = 85e8ccecccfbbcba1034cf1794a176fc80d8d973, deletion_comment = False)

yuslepukhin:
>         delete pEstimator;

If any of the above throws this does not execute.

---
Refers to: src/SharedLibrary/GeneratedCode/SharedLibrary_StandardScaleWrapperFeaturizer.cpp:69 in 85e8cce. [](commit_id = 85e8ccecccfbbcba1034cf1794a176fc80d8d973, deletion_comment = False)

yuslepukhin:
>         Microsoft::Featurizer::Featurizers::StandardScaleWrapperEstimator<std::int8_t> * pEstimator = g_pointerTable.Get<Microsoft::Featurizer::Featurizers::StandardScaleWrapperEstimator<std::int8_t>>(index);

Is there a method that both gets and removes the ptr in one step?

---
Refers to: src/SharedLibrary/GeneratedCode/SharedLibrary_StandardScaleWrapperFeaturizer.cpp:66 in 85e8cce. [](commit_id = 85e8ccecccfbbcba1034cf1794a176fc80d8d973, deletion_comment = False)

yuslepukhin:
>         size_t index = g_pointerTable.Add(pTransformer);

What happens to the newly created transformer if Add() throws?

---
Refers to: src/SharedLibrary/GeneratedCode/SharedLibrary_StandardScaleWrapperFeaturizer.cpp:235 in 85e8cce. [](commit_id = 85e8ccecccfbbcba1034cf1794a176fc80d8d973, deletion_comment = False)

yuslepukhin:
>         size_t index = g_pointerTable.Add(pTransformer);

Same comment as for Transsformer

---
Refers to: src/SharedLibrary/GeneratedCode/SharedLibrary_StandardScaleWrapperFeaturizer.cpp:261 in 85e8cce. [](commit_id = 85e8ccecccfbbcba1034cf1794a176fc80d8d973, deletion_comment = False)

yuslepukhin:
>         unsigned char * new_buffer(new unsigned char[buffer.size()]);

This should be ByteArray::value_type. Currently, uint8_t.

---
Refers to: src/SharedLibrary/GeneratedCode/SharedLibrary_StandardScaleWrapperFeaturizer.cpp:314 in 85e8cce. [](commit_id = 85e8ccecccfbbcba1034cf1794a176fc80d8d973, deletion_comment = False)

yuslepukhin:
>         std::copy(buffer.begin(), buffer.end(), new_buffer);

memcpy_s?

---
Refers to: src/SharedLibrary/GeneratedCode/SharedLibrary_StandardScaleWrapperFeaturizer.cpp:316 in 85e8cce. [](commit_id = 85e8ccecccfbbcba1034cf1794a176fc80d8d973, deletion_comment = False)

yuslepukhin:
>         Microsoft::Featurizer::Featurizers::StandardScaleWrapperEstimator<std::int8_t>::TransformerType & transformer(*g_pointerTable.Get<Microsoft::Featurizer::Featurizers::StandardScaleWrapperEstimator<std::int8_t>::TransformerType>(reinterpret_cast<size_t>(pHandle)));

This line requires more than one monitor to review.

---
Refers to: src/SharedLibrary/GeneratedCode/SharedLibrary_StandardScaleWrapperFeaturizer.cpp:342 in 85e8cce. [](commit_id = 85e8ccecccfbbcba1034cf1794a176fc80d8d973, deletion_comment = False)

src/SharedLibrary/IntegrationTests/SharedLibrary_SSWFeaturizer_IntegrationTest.cpp
```diff
@@ -0,0 +1,84 @@
...
+            static_cast<std::int8_t>(0),
+            static_cast<std::int8_t>(0),
+            static_cast<std::int8_t>(1),
+            static_cast<std::int8_t>(1)
+        },
+        std::vector<std::int8_t>{
+            static_cast<std::int8_t>(2)
+        },
+        [](std::vector<std::double_t> const &args) {
+            return args == std::vector<std::double_t>{static_cast<std::double_t>(3)};
```
yuslepukhin: >static_cast<std::double_t>(3) [](start = 54, length = 29)

would 3.0 be a bit more readable?

src/SharedLibrary/IntegrationTests/SharedLibrary_SSWFeaturizer_IntegrationTest.cpp
```diff
@@ -0,0 +1,84 @@
...
+}
+
+TEST_CASE("double_t_without_mean_with_std") {
+    StandardScaleWrapperFeaturizer_double_Test(
+        std::vector<std::double_t>{
+            static_cast<std::double_t>(0),
+            static_cast<std::double_t>(0),
+            static_cast<std::double_t>(1),
+            static_cast<std::double_t>(1)
+        },
```
yuslepukhin: { 0., 0., 1., 1. } is that is required.

src/SharedLibrary/IntegrationTests/SharedLibrary_SSWFeaturizer_IntegrationTest.cpp
angryjinyan: These two bools are configuration params for this featurizer which indicate should we use mean and standard deviation of input data.

src/Tools/CodeGenerator/Featurizers.yaml
```diff
@@ -634,15 +634,51 @@ featurizers:
...
+          - int8
+          - int16
+          - int32
+          - int64
+          - uint8
+          - uint16
+          - uint32
+          - uint64
+          - float
+          - double
```
yuslepukhin: nit: suggest to place these together sizewise. ORT convention is to place int8, uint8, int16, uint16 etc.

src/SharedLibrary/IntegrationTests/SharedLibrary_SSWFeaturizer_IntegrationTest.cpp
angryjinyan: 3.0 is a better way to write it.
=={3.0} is not a valid expression since initializer list cannot be used on the right hand side of operator '=='.

src/SharedLibrary/IntegrationTests/SharedLibrary_SSWFeaturizer_IntegrationTest.cpp
yuslepukhin: Do they stop being bool if you wrote true or false?

---
In reply to: [366533146](https://github.com/microsoft/FeaturizersLibrary/pull/65#discussion_r366533146) [](ancestors = 366533146)

src/SharedLibrary/IntegrationTests/SharedLibrary_SSWFeaturizer_IntegrationTest.cpp
yuslepukhin: Creating a container to compare a single element is not a good choice.

---
In reply to: [366543401](https://github.com/microsoft/FeaturizersLibrary/pull/65#discussion_r366543401) [](ancestors = 366543401)

src/Tools/CodeGenerator/Featurizers.yaml
angryjinyan: I think this is a better way to place these types. Will fix this for the whole file in a separate PR

PR #67, opened 2020-01-15
implement Traits<Matrix> serialization and deserialization

wangyems:
1, introduce eigen in this PR
2, include eigen in Traits.h
3, implement Traits<Eigen::matrixX<T>> method


src/Traits.h
```diff
@@ -872,6 +898,55 @@ struct Traits<std::unordered_map<KeyT, T, HashT, KeyEqualT, AllocatorT>> : publi
...
+        for (typename MatrixT::Index colId = 0; colId < value.cols(); ++colId) {
+            Traits<typename MatrixT::Scalar>::serialize(ar, value(rowId, colId));
+        }   
+    }
+    
+    return ar;
+}
+
+template <typename MatrixT, typename ArchiveT>
+MatrixT DeserializeMatrix(ArchiveT &ar) {
```
davidbrownellWork: Do these need to be defined in an anonymous namespace or can then be defined in line below?

src/Traits.h
```diff
@@ -872,6 +898,55 @@ struct Traits<std::unordered_map<KeyT, T, HashT, KeyEqualT, AllocatorT>> : publi
...
+    typename MatrixT::Index                 numRows(ar.template deserialize<typename MatrixT::Index>());
+    typename MatrixT::Index                 numCols(ar.template deserialize<typename MatrixT::Index>());
+
+    MatrixT                                 result(numRows, numCols);
+
+    for (typename MatrixT::Index rowId = 0; rowId < numRows; ++rowId) {
+        for (typename MatrixT::Index colId = 0; colId < numCols; ++colId) {
+            typename MatrixT::Scalar        element(Traits<typename MatrixT::Scalar>::deserialize(ar));
+            result(rowId, colId) = element;
```
davidbrownellWork: >element [](start = 35, length = 7)

Avoid a copy here by either:
A) using std::move
B) eliminating the use of the var and assigning result(rowId, colId) = Traits<typename MatrixT::Scalar>::deserialize(ar);

src/Traits.h
```diff
@@ -872,6 +898,55 @@ struct Traits<std::unordered_map<KeyT, T, HashT, KeyEqualT, AllocatorT>> : publi
...
+        }   
+    }
+
+    return result;
+}
+
+} // anonymous namespace
+
+template <typename T>
+struct Traits<Eigen::MatrixX<T>> {
```
davidbrownellWork: >Traits [](start = 7, length = 6)

We will need to add additional functionality to make this Trait class compliant with the others. See the list on line 155.

davidbrownellWork: >Traits [](start = 7, length = 6)

Add unit tests for all functionality.

PR #69, opened 2020-01-16
PCAFeaturizer

wangyems:


src/Featurizers/PCAFeaturizer.h
```diff
@@ -0,0 +1,267 @@
...
+
+    void save(Archive &ar) const override;
+
+private:
+    // ----------------------------------------------------------------------
+    // |
+    // |  Private Data
+    // |
+    // ----------------------------------------------------------------------
+    MatrixT const                                    _eigenvalues;
```
davidbrownellWork: >MatrixT const                                    _eigenvalues; [](start = 4, length = 62)

Is this being used?

src/Featurizers/PCAFeaturizer.h
```diff
@@ -0,0 +1,267 @@
...
+    // |
+    // |  Private Methods
+    // |
+    // ----------------------------------------------------------------------
+     // MSVC has problems when the declaration and definition are separated
+    void execute_impl(typename BaseType::InputType const &input, typename BaseType::CallbackFunction const &callback) override {
+        if (input.size() == 0)
+            throw std::invalid_argument("Input matrix size() invalid");
+
+        if (input.cols() != _eigenvectors.cols())
```
davidbrownellWork: >if (input.cols() != _eigenvectors.cols()) [](start = 8, length = 41)

Can you add a comment if this code is correct? It seems strange to me that it expects the col values to match.

src/Featurizers/PCAFeaturizer.h
```diff
@@ -0,0 +1,267 @@
...
+        std::move(
+            [this, &colIndex](void) -> size_t & {
+                if(colIndex >= this->get_column_annotations().size())
+                    throw std::invalid_argument("colIndex");
+
+                return colIndex;
+            }()
+        )
+    ) {
+        _hasCalledFit = false;
```
davidbrownellWork: >_hasCalledFit [](start = 8, length = 13)

Initialize with the other member vars rather than inside the body of the constructor.

src/Featurizers/PCAFeaturizer.h
```diff
@@ -0,0 +1,267 @@
...
+    // ----------------------------------------------------------------------
+    // |
+    // |  Private Methods
+    // |
+    // ----------------------------------------------------------------------
+    bool begin_training_impl(void) override;
+    // MSVC has problems when the declaration and definition are separated
+    FitResult fit_impl(typename BaseType::InputType const *pBuffer, size_t) override {
+        if (_hasCalledFit)
+            throw std::runtime_error("Fit_impl() should not be called move than once in PCAFeaturizer");
```
davidbrownellWork: >Fit_impl [](start = 38, length = 8)

fit_impl (change case)

src/Featurizers/PCAFeaturizer.h
```diff
@@ -0,0 +1,267 @@
...
+            return cov;
+        }()
+    );
+
+    // free _matrix
+    _matrix.resize(0, 0);
+
+    //PCA uses sigma but sklearn PCA uses sqrt(sigma), this implementation follows original principle.
+    this->_eigenvalues = eig.eigenvalues();
+    this->_eigenvectors = eig.eigenvectors();
```
davidbrownellWork: Is it better to invoke this functionality here or when the transformer is created? If you move it to create_transformer_impl, you can remove _eigenvalues and _eigenvectors as members of the class.

src/Featurizers/PCAFeaturizer.h
```diff
@@ -0,0 +1,267 @@
...
+template <typename MatrixT,size_t MaxNumTrainingItemsV>
+bool PCAEstimator<MatrixT, MaxNumTrainingItemsV>::begin_training_impl(void) /*override*/ {
+    return true;
+}
+
+template <typename MatrixT, size_t MaxNumTrainingItemsV>
+void PCAEstimator<MatrixT, MaxNumTrainingItemsV>::complete_training_impl(void) /*override*/ {
+
+    Eigen::SelfAdjointEigenSolver<MatrixT> eig(
+        [&]() -> MatrixT {
```
davidbrownellWork: >& [](start = 9, length = 1)

Make what you are capturing explicit. Maybe you want to capture `this`?

src/Featurizers/PCAFeaturizer.h
wangyems: Not for transformer but for potential annotation(eigenvalues) retrieving

src/Featurizers/PCAFeaturizer.h
wangyems: because this is simplified(mathematically) from input.col() == _eigenvectors.transpose().rows(). I 'll add comments for it

src/Featurizers/PCAFeaturizer.h
wangyems: sure

src/Featurizers/PCAFeaturizer.h
wangyems: sure

src/Featurizers/PCAFeaturizer.h
davidbrownellWork: If that is the case, create an Estimator in Components and have this Estimator use that one. Because of how close we are to release, I'd recommend removing the value rather than going through a refactor like this.

wangyems: ok

PR #72, opened 2020-01-16
Normalizer

angryjinyan:
Finish Vector norm estimator and NormalizeFeaturizer

Will add more unit tests for the featurizer


yuslepukhin:
> #   pragma clang diagnostic ignored "-Wfloat-equal"

@David this is wrong. The warning should not be suppressed.

---
Refers to: src/Featurizers/Structs.h:283 in ae08450. [](commit_id = ae084502df2faaa3187ea88ba52cc6d7b7fafe41, deletion_comment = False)

wangyems:
> > # pragma clang diagnostic ignored "-Wfloat-equal"
> 
> @david this is wrong. The warning should not be suppressed.
> 
> Refers to: src/Featurizers/Structs.h:283 in [ae08450](https://github.com/microsoft/FeaturizersLibrary/commit/ae084502df2faaa3187ea88ba52cc6d7b7fafe41). [](commit_id = [ae08450](https://github.com/microsoft/FeaturizersLibrary/commit/ae084502df2faaa3187ea88ba52cc6d7b7fafe41), deletion_comment = False)

@davidbrownellWork

src/Featurizers/NormalizeFeaturizer.h
```diff
@@ -0,0 +1,325 @@
...
+
+    using ValueType                         = typename InputT::value_type;
+    using BaseType                          = StandardTransformer<InputT, std::vector<long double>>;
+
+    // ----------------------------------------------------------------------
+    // |
+    // |  Public Methods
+    // |
+    // ----------------------------------------------------------------------
+    NormalizeTransformer(std::vector<std::double_t> norms);
```
yuslepukhin: >  No [](start = 2, length = 4)

explicit?

src/Featurizers/NormalizeFeaturizer.h
```diff
@@ -0,0 +1,325 @@
...
+    NormalizeTransformer(Archive &ar);
```
yuslepukhin: >   [](start = 2, length = 2)

explicit?

src/Featurizers/NormalizeFeaturizer.h
```diff
@@ -0,0 +1,325 @@
...
+// ----------------------------------------------------------------------
+// |
+// |  NormalizeTransformer
+// |
+// ----------------------------------------------------------------------
+template <typename InputT>
+NormalizeTransformer<InputT>::NormalizeTransformer(std::vector<std::double_t> norms) :
+    _norms(std::move(norms)),
+    _row(0) {
+        for (auto it = norms.begin(); it != norms.end(); ++it) {
```
yuslepukhin: >or (auto it = norms.begin(); it != norms.end(); ++it) { [](start = 9, length = 55)

range loop like for(norm : norms) {}

src/Traits.h
```diff
@@ -207,7 +207,7 @@ template <typename T>
 struct TraitsImpl {
     static constexpr bool const             IsNullableType = false;
     static constexpr bool const             IsNativeNullableType = false;
-
+    static constexpr bool const             IsVectorType = false;
```
davidbrownellWork: >static constexpr bool const             IsVectorType = false; [](start = 4, length = 61)

This represents a slippery slope - if we add this,  should we add something that detects maps? Smart pointers? etc. This code should probably remain local to where it is used rather than putting it in Traits.h. 

If it needs to remain in Traits.h, there should be corresponding updates in Traits_UnitTest.cpp that validates the functionality.

davidbrownellWork:
> #   pragma clang diagnostic ignored "-Wfloat-equal"

This suppression is happening in a number of different files. If we want to remove it, let's introduce a convention that isn't too onerous (for example, I'm hoping that we don't need to introduce the concept of an epsilon as the comparisons that trigger this warning are generally within the context of very simple unit tests that don't necessitate that level or precision).

---
In reply to: [575390200](https://github.com/microsoft/FeaturizersLibrary/pull/72#issuecomment-575390200) [](ancestors = 575390200)

---
Refers to: src/Featurizers/Structs.h:283 in ae08450. [](commit_id = ae084502df2faaa3187ea88ba52cc6d7b7fafe41, deletion_comment = False)

src/Featurizers/Structs.h
```diff
@@ -95,6 +95,65 @@ class SingleValueSparseVectorEncoding {
...
+
+    bool operator==(CSREncoding const &other) const;
+    bool operator!=(CSREncoding const &other) const;
+};
+
+/////////////////////////////////////////////////////////////////////////
+///  \enum          NormType
+///  \brief         Training state associated with an `Estimator`.
+///
+enum class NormType : unsigned char {
```
davidbrownellWork: >NormType [](start = 11, length = 8)

This doesn't belong in this file - move this definition closer to where it is actually used.

src/Featurizers/Structs.h
```diff
@@ -95,6 +95,65 @@ class SingleValueSparseVectorEncoding {
...
+    std::vector<value_type>          Values;
+    std::vector<std::uint64_t>       ColIndex;
+    std::vector<std::uint64_t>       RowIndex;
+
+    // ----------------------------------------------------------------------
+    // |
+    // |  Public Methods
+    // |
+    // ----------------------------------------------------------------------
+    CSREncoding();
```
davidbrownellWork: >CSREncoding(); [](start = 4, length = 14)

Remove this default constructor, as it creates the possibility of creating a CSREncoding object that doesn't have all of the values set. For example, is it valid to have a CSREncoding object with values but no ColIndex or RowIndexes? If this is to handle the creation of an empty matrix, create a static member named something like "CreateEmptyMatrix" that passes in empty vectors for Values, ColIndex, and RowIndex.

src/Featurizers/Structs.h
```diff
@@ -95,6 +95,65 @@ class SingleValueSparseVectorEncoding {
...
+
+    // ----------------------------------------------------------------------
+    // |
+    // |  Public Data
+    // |
+    // ----------------------------------------------------------------------
+
+    std::vector<value_type>          Values;
+    std::vector<std::uint64_t>       ColIndex;
+    std::vector<std::uint64_t>       RowIndex;
```
davidbrownellWork: These should be const.

src/Featurizers/Structs.h
```diff
@@ -95,6 +95,65 @@ class SingleValueSparseVectorEncoding {
...
+///                 CSREncoding contains three vectors to express a
+///                 sparse matrix:
+///                 V contains all non-zero element values,
+///                 ColIndex contains column indices of non-zero values,
+///                 RowIndex has one element per row in the matrix and encodes
+///                 the index in V where the given row starts
+///
+
+template <typename T>
+class CSREncoding {
```
davidbrownellWork: >CSREncoding [](start = 6, length = 11)

Suppress unintentional copies of this object with the FEATURIZER_MOVE_CONSTRUCTOR_ONLY. While you are at it, can you add TODO comments in SingleValueSparseVectorEncoding and SparseVectorEncoding to add this as well.

src/Featurizers/Structs.h
```diff
@@ -240,6 +299,56 @@ bool SingleValueSparseVectorEncoding<T>::operator!=(SingleValueSparseVectorEncod
...
 
+
+// ----------------------------------------------------------------------
+// |
+// |  CSREncoding
+// |
+// ----------------------------------------------------------------------
+template <typename T>
+CSREncoding<T>::CSREncoding() {
+    RowIndex.push_back(0);
```
davidbrownellWork: >RowIndex.push_back(0); [](start = 4, length = 22)

This should be implemented in terms of the functionality below - it should participate in the same validation logic.

src/Featurizers/Structs.h
```diff
@@ -240,6 +299,56 @@ bool SingleValueSparseVectorEncoding<T>::operator!=(SingleValueSparseVectorEncod
...
+}
+
+template <typename T>
+CSREncoding<T>::CSREncoding(std::vector<value_type> values, std::vector<std::uint64_t> col, std::vector<std::uint64_t> row) :
+    Values(std::move(values)),
+    ColIndex(std::move(col)),
+    RowIndex(std::move(row)) {
+        if(RowIndex[0] != 0) {
```
davidbrownellWork: >if(RowIndex[0] != 0) { [](start = 8, length = 22)

Will this statement be true for a matrix that doesn't have any values in the first row?

src/Featurizers/Components/NormUpdaters.h
```diff
@@ -67,6 +67,8 @@ class L1NormUpdater {
     // ----------------------------------------------------------------------
     L1NormUpdater(void);
 
+    void reset(void);
```
davidbrownellWork: >void reset(void); [](start = 4, length = 17)

I'd expect to see unit test changes that exercise this functionality.

src/Featurizers/Components/NormUpdaters.h
```diff
@@ -197,6 +203,11 @@ Updaters::L1NormUpdater<T>::L1NormUpdater() :
     _update_flag(false) {
```
davidbrownellWork: This logic is duplicated in reset. Perhaps call that method here?

davidbrownellWork:

This file would be a good place to store the enum that currently lives in Structs.h.

---
Refers to: src/Featurizers/Components/NormUpdaters.h:1 in 7801dad. [](commit_id = 7801dad2e48bdd63087993f11745e05160222a17, deletion_comment = False)

src/Featurizers/Components/VectorNormsEstimator.h
```diff
@@ -0,0 +1,181 @@
...
+// |  Details::VectorNormsTrainingOnlyPolicy
+// |
+// ----------------------------------------------------------------------
+template <typename T, NormType NormT>
+Details::VectorNormsTrainingOnlyPolicy<T, NormT>::VectorNormsTrainingOnlyPolicy(void) {
+}
+
+template <typename T, NormType NormT>
+void Details::VectorNormsTrainingOnlyPolicy<T, NormT>::fit(InputType const &input) {
+    for (auto it = input.begin(); it != input.end(); ++it) {
```
davidbrownellWork: >for (auto it = input.begin(); it != input.end(); ++it) [](start = 4, length = 54)

You don't need to use an iterator here, as you are immediately dereferencing it. For improved readability, change this to `for(auto const &value : input)`

src/Featurizers/Components/VectorNormsEstimator.h
```diff
@@ -0,0 +1,181 @@
...
+
+// ----------------------------------------------------------------------
+// |
+// |  VectorNormsAnnotationData
+// |
+// ----------------------------------------------------------------------
+VectorNormsAnnotationData::VectorNormsAnnotationData(std::vector<std::double_t> norms, NormType type) :
+    Norms(std::move(norms)),
+    Type(std::move(type)) {
+        for (auto it = norms.begin(); it != norms.end(); ++it) {
```
davidbrownellWork: >for (auto it = norms.begin(); it != norms.end(); ++it) [](start = 8, length = 54)

You don't need to use an iterator here, as you are immediately dereferencing it. For improved readability, change this to for(auto const &value : norms)

src/Featurizers/Components/VectorNormsEstimator.h
```diff
@@ -0,0 +1,181 @@
...
+namespace Featurizers {
+namespace Components {
+
+static constexpr char const * const         VectorNormsEstimatorName("VectorNormsEstimator");
+
+namespace TypeSelector {
+
+// updater type selector is created to select different updater based on input parameter
+template <typename T, NormType NormT>
+struct UpdaterTypeSelector {
```
davidbrownellWork: >UpdaterTypeSelector [](start = 7, length = 19)

I've made changes in StatisticalMetricsEstimator to make it more usable within other Featurizers that are similar to these. Take a look at those changes and try to incorporate those concepts here.

src/Featurizers/Components/VectorNormsEstimator.h
```diff
@@ -0,0 +1,181 @@
...
+///
+template <typename T, NormType NormT>
+class VectorNormsTrainingOnlyPolicy {
+public:
+    // ----------------------------------------------------------------------
+    // |
+    // |  Public Types
+    // |
+    // ----------------------------------------------------------------------
+    static_assert(Traits<T>::IsVectorType, "Input type has to be vector!");
```
davidbrownellWork: >static_assert(Traits<T>::IsVectorType, "Input type has to be vector!"); [](start = 4, length = 71)

We had discussed using iterators representing a range rather than iterator types; what was the result of that conversation?

src/Featurizers/Components/VectorNormsEstimator.h
```diff
@@ -0,0 +1,181 @@
...
+template <typename T, NormType NormT>
+Details::VectorNormsTrainingOnlyPolicy<T, NormT>::VectorNormsTrainingOnlyPolicy(void) {
+}
+
+template <typename T, NormType NormT>
+void Details::VectorNormsTrainingOnlyPolicy<T, NormT>::fit(InputType const &input) {
+    for (auto it = input.begin(); it != input.end(); ++it) {
+        _updater.update(*it);
+    }
+    _norms.push_back(static_cast<std::double_t>(_updater.commit()));
```
davidbrownellWork: >push_back [](start = 11, length = 9)

`emplace_back`

src/Featurizers/Components/VectorNormsEstimator.h
```diff
@@ -0,0 +1,181 @@
...
+///
+class VectorNormsAnnotationData {
+public:
+    // ----------------------------------------------------------------------
+    // |
+    // |  Public Data
+    // |
+    // ----------------------------------------------------------------------
+    std::vector<std::double_t> const Norms;
+    NormType                 const Type;
```
davidbrownellWork: >NormType                 const Type; [](start = 4, length = 36)

This should be a template rather than a data member, as we want to be able to store annotation data for both a L1 and L2 norm simultaneously. As implemented, get_annotation_data will return a cached L1 norm even when a L2 norm is requested (as it won't be able to resolve that those are 2 different things).

src/Featurizers/NormalizeFeaturizer.h
davidbrownellWork: >norms [](start = 23, length = 5)

norms will be invalid here, as it has been moved to _norms. You will want to operate only on _norms.

src/Featurizers/NormalizeFeaturizer.h
```diff
@@ -0,0 +1,325 @@
...
+// ----------------------------------------------------------------------
+
+// ----------------------------------------------------------------------
+// |
+// |  NormalizeTransformer
+// |
+// ----------------------------------------------------------------------
+template <typename InputT>
+NormalizeTransformer<InputT>::NormalizeTransformer(std::vector<std::double_t> norms) :
+    _norms(std::move(norms)),
```
davidbrownellWork: >_norms [](start = 4, length = 6)

Is it valid to have an empty norms?

src/Featurizers/NormalizeFeaturizer.h
```diff
@@ -0,0 +1,325 @@
...
+                throw std::invalid_argument("Norms shouldn't be less than 0!");
+            }
+        }
+}
+
+template <typename InputT>
+NormalizeTransformer<InputT>::NormalizeTransformer(Archive &ar) :
+    NormalizeTransformer(
+        [&ar](void) {
+            std::vector<std::double_t> norms(Traits<std::vector<std::double_t>>::deserialize(ar));
```
davidbrownellWork: >std::vector<std::double_t> norms(Traits<std::vector<std::double_t>>::deserialize(ar)); [](start = 12, length = 86)

Deserialize and validate the major and minor version.

src/Featurizers/NormalizeFeaturizer.h
```diff
@@ -0,0 +1,325 @@
...
+
+            return NormalizeTransformer<InputT>(std::move(norms));
+        }()
+    ) {
+}
+
+template <typename InputT>
+void NormalizeTransformer<InputT>::save(Archive &ar) const /*override*/ {
+    Traits<std::vector<std::double_t>>::serialize(ar, _norms);
```
davidbrownellWork: >Traits<std::vector<std::double_t>>::serialize(ar, _norms); [](start = 4, length = 58)

Serialize the major and minor version (see other featurizers for an example of this).

src/Featurizers/NormalizeFeaturizer.h
```diff
@@ -0,0 +1,325 @@
...
+    _row(0) {
+        for (auto it = norms.begin(); it != norms.end(); ++it) {
+            if (*it < 0) {
+                throw std::invalid_argument("Norms shouldn't be less than 0!");
+            }
+        }
+}
+
+template <typename InputT>
+NormalizeTransformer<InputT>::NormalizeTransformer(Archive &ar) :
```
davidbrownellWork: >NormalizeTransformer<InputT>::NormalizeTransformer(Archive &ar) [](start = 0, length = 63)

Add a unit test that throws an error when invalid version numbers are deserialized (see the other featurizers for an example of this).

src/Featurizers/NormalizeFeaturizer.h
```diff
@@ -0,0 +1,325 @@
...
+    execute_impl(input, callback, std::integral_constant<bool, Microsoft::Featurizer::Traits<ValueType>::IsNullableType>());
+}
+
+template <typename InputT>
+void NormalizeTransformer<InputT>::execute_impl(typename BaseType::InputType const &input, typename BaseType::CallbackFunction const &callback, std::true_type) {
+    // ----------------------------------------------------------------------
+    using InputTraits                       = Traits<ValueType>;
+    using TransformedTraits                 = Traits<long double>;
+    // ----------------------------------------------------------------------
+    std::vector<long double> res;
```
davidbrownellWork: >res [](start = 29, length = 3)

Since you know what the size of this vector is going to be, prereserve the space via `res.reserve(input.size());`

src/Featurizers/NormalizeFeaturizer.h
```diff
@@ -0,0 +1,325 @@
...
+    for (auto it = input.begin(); it != input.end(); ++it) {
+        if(InputTraits::IsNull(*it)) {
+            res.push_back(TransformedTraits::CreateNullValue());
+            continue;
+        }
+        res.push_back(static_cast<long double>(InputTraits::GetNullableValue(*it)) / static_cast<long double>(_norms[_row]));
+    }
+    ++_row;
+    callback(res);
```
davidbrownellWork: >res [](start = 13, length = 3)

Why is this returning a vector of results? What happens if this method is called multiple times and _row becomes greater than _norms? This doesn't feel right - let's discuss.

src/Featurizers/NormalizeFeaturizer.h
```diff
@@ -0,0 +1,325 @@
...
+}
+
+template <typename InputT>
+void NormalizeTransformer<InputT>::execute_impl(typename BaseType::InputType const &input, typename BaseType::CallbackFunction const &callback, std::false_type) {
+    std::vector<long double> res;
+    for (auto it = input.begin(); it != input.end(); ++it) {
+        res.push_back(static_cast<long double>(*it) / static_cast<long double>(_norms[_row]));
```
davidbrownellWork: >push_back [](start = 12, length = 9)

`emplace_back`

src/Featurizers/UnitTests/NormalizeFeaturizer_UnitTests.cpp
```diff
@@ -0,0 +1,41 @@
+// ----------------------------------------------------------------------
```
davidbrownellWork: I expect to see more tests here - at least 1 test for every exception thrown. There seems to be more functionality that this should exercise as well.

src/Traits.h
angryjinyan: Input type is no longer a vector. IsVectorType removed

src/Featurizers/Structs.h
angryjinyan: Eigen library can deal with compressed format so CSREncoding is also no longer necessary

src/Featurizers/Structs.h
angryjinyan: FEATURIZER_MOVE_CONSTRUCTOR_ONLY added

src/Featurizers/NormalizeFeaturizer.h
```diff
@@ -0,0 +1,325 @@
+// ----------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License
+// ----------------------------------------------------------------------
+#pragma once
```
yuslepukhin: >#pragma on [](start = 0, length = 10)

This should be added to cmake file

src/Featurizers/Components/VectorNormsEstimator.h
angryjinyan: I don't think it applies here since function names such as reset, update and commit are the same for all three norm updaters and I just need to select appropriate updater type regarding NormType

angryjinyan: Resolved after discussion

src/Featurizers/TestHelpers.h
```diff
@@ -138,6 +138,45 @@ bool FuzzyCheck(std::vector<T> const & vec1, std::vector<T> const & vec2, std::d
     }
     return true;
 }
+template<>
```
davidbrownellWork: We don't need to use a template here, as functional overloads will work just fine.

src/Featurizers/TestHelpers.h
```diff
@@ -138,6 +138,45 @@ bool FuzzyCheck(std::vector<T> const & vec1, std::vector<T> const & vec2, std::d
...
+            return false;
+        }
+        if (abs(vec1[idx] - vec2[idx]) > static_cast<std::float_t>(epsilon)) {
+            return false;
+        }
+    }
+    return true;
+}
+template <>
+bool FuzzyCheck<std::double_t>(std::vector<std::double_t> const & vec1, std::vector<std::double_t> const & vec2, std::double_t epsilon) {
```
davidbrownellWork: >bool [](start = 0, length = 4)

This is the same logic as above. Use the template to eliminate the duplicated logic.

src/Featurizers/Components/NormUpdaters.h
```diff
@@ -1,298 +1,373 @@
...
+template <typename T>
+class L1NormUpdater {
+public:
+
+    // ----------------------------------------------------------------------
+    // |
+    // |  Public Types
+    // |
+    // ----------------------------------------------------------------------
+    using InputType                         = T;
```
davidbrownellWork: >using InputType                         = T; [](start = 4, length = 44)

InputType isn't a concept that directly applies to this class - T should be sufficient.

src/Featurizers/Components/NormUpdaters.h
```diff
@@ -1,298 +1,373 @@
...
-        throw std::runtime_error("Overflows occured during calculating l1_norm! Check your data!");
-    }
-}
-
-template <typename T>
-long double Updaters::L1NormUpdater<T>::commit(void) {
-    assert(_l1_norm >= 0);
-    if (!_update_flag) {
-        throw std::runtime_error("update is not called before l1 is committed!");
-    }
```
davidbrownellWork: By removing this clause, it is possible that commit can be called when no valid input is provided. I think that this should be restored, as that is an error condition.

src/Featurizers/Components/VectorNormsEstimator.h
```diff
@@ -0,0 +1,167 @@
...
+    FEATURIZER_MOVE_CONSTRUCTOR_ONLY(VectorNormsAnnotationData);
+};
+
+namespace Details {
+
+/////////////////////////////////////////////////////////////////////////
+///  \class         VectorNormsTrainingOnlyPolicy
+///  \brief         `VectorNormsEstimator` implementation details.
+///
+template <typename T, typename UpdaterType>
```
davidbrownellWork: >T [](start = 19, length = 1)

Rename this to IteratorRangeT to make it clear what is expected. Also, static_assert that this type is a tuple or pair.

src/Featurizers/Components/VectorNormsEstimator.h
```diff
@@ -0,0 +1,167 @@
...
+class VectorNormsTrainingOnlyPolicy {
+public:
+    // ----------------------------------------------------------------------
+    // |
+    // |  Public Types
+    // |
+    // ----------------------------------------------------------------------
+
+    using IteratorType = typename std::tuple_element<0, T>::type;
+    using ValueType    = typename std::iterator_traits<IteratorType>::value_type;
```
davidbrownellWork: These types aren't used in the public declarations, so they should be private.

src/Featurizers/Components/VectorNormsEstimator.h
davidbrownellWork: >UpdaterType [](start = 31, length = 11)

Rename to UpdaterT

src/Featurizers/Components/VectorNormsEstimator.h
```diff
@@ -0,0 +1,167 @@
...
+// ----------------------------------------------------------------------
+
+// ----------------------------------------------------------------------
+// |
+// |  VectorNormsAnnotationData
+// |
+// ----------------------------------------------------------------------
+template <typename UpdaterType>
+VectorNormsAnnotationData<UpdaterType>::VectorNormsAnnotationData(std::vector<std::double_t> norms) :
+    Norms(std::move(norms)) {
```
davidbrownellWork: >Norms(std::move(norms)) [](start = 4, length = 23)

Is it valid to have an empty vector of norms?

src/Featurizers/Components/VectorNormsEstimator.h
```diff
@@ -0,0 +1,167 @@
...
+// ----------------------------------------------------------------------
+template <typename T, typename UpdaterType>
+Details::VectorNormsTrainingOnlyPolicy<T, UpdaterType>::VectorNormsTrainingOnlyPolicy(void) {
+}
+
+template <typename T, typename UpdaterType>
+void Details::VectorNormsTrainingOnlyPolicy<T, UpdaterType>::fit(InputType const &input) {
+    // T is a pair of iterators
+    // unpack input to two iterators
+    IteratorType & begin = const_cast<IteratorType&>(std::get<0>(input));
```
davidbrownellWork: >const_cast<IteratorType&> [](start = 27, length = 25)

This const_cast shouldn't be necessary. If you need to use the value for iteration, make a copy rather than taking a reference.

src/Featurizers/NormalizeFeaturizer.h
```diff
@@ -0,0 +1,362 @@
...
+#include "Structs.h"
+namespace Microsoft {
+namespace Featurizer {
+namespace Featurizers {
+
+/////////////////////////////////////////////////////////////////////////
+///  \class         NormalizeTransformer
+///  \brief         Normalize based on l1, l2 or max norm
+///
+template <typename InputT>
```
davidbrownellWork: >InputT [](start = 19, length = 6)

IteratorRangeT; static_assert that this type is a 2-element tuple or pair.

src/Featurizers/NormalizeFeaturizer.h
```diff
@@ -0,0 +1,362 @@
...
+public:
+    // ----------------------------------------------------------------------
+    // |
+    // |  Public Types
+    // |
+    // ----------------------------------------------------------------------
+
+    
+    using IteratorType = typename std::tuple_element<0, InputT>::type;
+    using ValueType    = typename std::iterator_traits<IteratorType>::value_type;
```
davidbrownellWork: These types aren't used in public declarations so make them private.

src/Featurizers/NormalizeFeaturizer.h
```diff
@@ -0,0 +1,362 @@
...
+// |
+// ----------------------------------------------------------------------
+template <typename InputT>
+NormalizeTransformer<InputT>::NormalizeTransformer(std::vector<std::double_t> norms) :
+    _norms(std::move(norms)),
+    _row(0) {
+        if (_norms.empty()) {
+            throw std::invalid_argument("Norms input to NormalizeTransformer is empty!");
+        }
+        for (auto norm : _norms) {
```
davidbrownellWork: >auto [](start = 13, length = 4)

This is copying the value; should this be `auto const &`?

src/Featurizers/NormalizeFeaturizer.h
```diff
@@ -0,0 +1,362 @@
...
+
+    std::vector<std::double_t> res;
+    res.reserve(static_cast<size_t>(std::distance(begin, end)));
+    while (begin != end) {
+        if(InputTraits::IsNull(*begin)) {
+            res.emplace_back(TransformedTraits::CreateNullValue());
+            ++begin;
+            continue;
+        }
+        res.emplace_back(static_cast<std::double_t>(static_cast<long double>(InputTraits::GetNullableValue(*begin)) / static_cast<long double>(_norms[_row])));
```
davidbrownellWork: Rather than embedding the continue in the if clause, I'd recommend removing the code that increments the iterator and continue and adding an else before the other emplace. Doing this better encapsulates what the code is doing: adding a value to the results and then moving the iterator to point to the next item.

src/Featurizers/UnitTests/NormalizeFeaturizer_UnitTests.cpp
davidbrownellWork: Can you add a test for no inputs and inputs that are all null.

---
In reply to: [368049067](https://github.com/microsoft/FeaturizersLibrary/pull/72#discussion_r368049067) [](ancestors = 368049067)

src/Featurizers/NormalizeFeaturizer.h
```diff
@@ -0,0 +1,373 @@
...
+void NormalizeTransformer<IteratorRangeT>::execute_impl(IteratorRangeT const &input, typename BaseType::CallbackFunction const &callback) /*override*/ {
+    // unpack input to two iterators
+    IteratorType begin = std::get<0>(input);
+    IteratorType const & end = std::get<1>(input);
+
+    execute_impl(begin, end, callback, std::integral_constant<bool, Microsoft::Featurizer::Traits<ValueType>::IsNullableType>());
+}
+
+template <typename IteratorRangeT>
+void NormalizeTransformer<IteratorRangeT>::execute_impl(IteratorType &begin, IteratorType const &end, typename BaseType::CallbackFunction const &callback, std::true_type) {
```
davidbrownellWork: >NormalizeTransformer [](start = 5, length = 20)

The code in these 2 methods is almost identical, with the only difference being how the iterator is accessed (is it wrapped in GetNullableValue or not). Can you think of a way to implement this logic in a single function that can be called by both methods?

src/Featurizers/Components/NormUpdaters.h
```diff
@@ -1,298 +1,381 @@
...
+
+template <typename T>
+void Updaters::L1NormUpdater<T>::update_impl(T input, std::false_type /*is_nullable*/) {
+    update_impl(input);
+}
+
+template <typename T>
+template <typename U>
+void Updaters::L1NormUpdater<T>::update_impl(U input) {
+    if (!_has_valid_input)
```
davidbrownellWork: >if (!_has_valid_input) [](start = 4, length = 22)

You can remove this if check and always set the value to true (once it is true, it will always be true).

src/Featurizers/Components/NormUpdaters.h
```diff
@@ -1,298 +1,381 @@
...
+    // check if exceeds bound
+    // only check for l2_norm and no checks for l1_norm is because if l1_norm is out of upper bound, l2_norm is definitely out of bound
+    if (std::isinf(_l1_norm)) {
+        throw std::runtime_error("Overflows occured during calculating l1_norm! Check your data!");
+    }
+}
+
+template <typename T>
+long double Updaters::L1NormUpdater<T>::commit(void) {
+    assert(_l1_norm >= 0);
```
davidbrownellWork: >assert(_l1_norm >= 0); [](start = 4, length = 22)

Nit: move this below the if check. We shouldn't look at _l1_norm until we know that it is valid.

src/Featurizers/NormalizeFeaturizer.h
```diff
@@ -0,0 +1,373 @@
...
+} // namespace Details
+
+/////////////////////////////////////////////////////////////////////////
+///  \class         NormalizeEstimator
+///  \brief         Chains up VectorNormsEstimator and NormalizeTransformerImpl
+///                 to normalize input vector data so that it's norm becoming 1
+///
+template <
+    typename IteratorRangeT,
+    typename UpdaterType,
```
davidbrownellWork: >typename UpdaterType, [](start = 4, length = 21)

The code gen code won't be able to instantiate this Estimator, as it won't be able to determine what this value should be. If you need to distinguish between L1 and L2 functionality, you might be looking at 2 different Estimators.

src/Featurizers/Components/VectorNormsEstimator.h
```diff
@@ -0,0 +1,192 @@
...
+template <typename T>
+struct IsIteratorPair<std::pair<T, T>> {
+    static constexpr bool const value = true;
+};
+
+/////////////////////////////////////////////////////////////////////////
+///  \class         VectorNormsAnnotationData
+///  \brief         An annotation class which contains the norms of a matrix
+///
+template <typename UpdaterT>
```
davidbrownellWork: This template definition is going to be problematic. Let's discuss in person.

src/Featurizers/NormalizeFeaturizer.h
```diff
@@ -0,0 +1,373 @@
+// ----------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
```
davidbrownellWork: Add this file to cmake/FeaturizersCode.cmake

src/Featurizers/Components/VectorNormsEstimator.h
```diff
@@ -0,0 +1,192 @@
+// ----------------------------------------------------------------------
```
davidbrownellWork: Add this file to cmake/FeaturizersCode.cmake

src/Featurizers/Components/NormUpdaters.h
angryjinyan: Good catch!

src/Featurizers/NormalizeFeaturizer.h
```diff
@@ -0,0 +1,417 @@
...
+///  \class         NormalizeEstimator
+///  \brief         Chains up VectorNormsEstimator and NormalizeTransformerImpl
+///                 to normalize input vector data so that it's norm becoming 1
+///
+template <
+    typename IteratorRangeT,
+    typename UpdaterT,
+    size_t MaxNumTrainingItemsV=std::numeric_limits<size_t>::max()
+>
+class NormalizeEstimator :
```
davidbrownellWork: >NormalizeEstimator [](start = 6, length = 18)

Nit: decorate this name to make it clear that it shouldn't be used in isolation; maybe `NormalizeEstimatorImpl`?

src/Featurizers/NormalizeFeaturizer.h
```diff
@@ -0,0 +1,417 @@
...
+    return Traits<ValueType>::IsNull(*input);
+}
+template <typename IteratorRangeT>
+bool NormalizeTransformer<IteratorRangeT>::is_null(IteratorType const &, std::false_type) {
+    return false;
+}
+
+
+template <typename IteratorRangeT>
+typename GetNullableType<typename std::iterator_traits<typename std::tuple_element<0, IteratorRangeT>::type>::value_type>::type NormalizeTransformer<IteratorRangeT>::get_value(IteratorType const &input, std::true_type) {
```
davidbrownellWork: >typename GetNullableType<typename std::iterator_traits<typename std::tuple_element<0, IteratorRangeT>::type>::value_type>::type [](start = 0, length = 127)

This should be `typename GetNullableType<typename std::iterator_traits<typename std::tuple_element<0, IteratorRangeT>::type>::value_type>::type const &` to avoid a copy.

src/Featurizers/NormalizeFeaturizer.h
```diff
@@ -0,0 +1,417 @@
...
+    return Traits<ValueType>::GetNullableValue(*input);
+}
+template <typename IteratorRangeT>
+typename std::iterator_traits<typename std::tuple_element<0, IteratorRangeT>::type>::value_type NormalizeTransformer<IteratorRangeT>::get_value(IteratorType const &input, std::false_type) {
```
davidbrownellWork: >typename std::iterator_traits<typename std::tuple_element<0, IteratorRangeT>::type>::value_type [](start = 0, length = 95)

This should be `typename GetNullableType<typename std::iterator_traits<typename std::tuple_element<0, IteratorRangeT>::type>::value_type>::type const &` to avoid a copy.

src/Featurizers/NormalizeFeaturizer.h
```diff
@@ -0,0 +1,417 @@
...
+    void execute_impl(IteratorRangeT const &input, typename BaseType::CallbackFunction const &callback) override;
+
+    void execute_impl(IteratorType &begin, IteratorType const &end, typename BaseType::CallbackFunction const &callback, std::true_type);
+    void execute_impl(IteratorType &begin, IteratorType const &end, typename BaseType::CallbackFunction const &callback, std::false_type);
+
+    bool is_null(IteratorType const &input, std::true_type);
+    bool is_null(IteratorType const &input, std::false_type);
+
+    typename GetNullableType<ValueType>::type get_value(IteratorType const &input, std::true_type);
+    ValueType get_value(IteratorType const &input, std::false_type);
```
davidbrownellWork: These functions aren't mutating data, so they should be `const`, however...
These functions aren't using any class data, so they should be `static`. however...
These functions aren't adding functionality to the class, so they should be free functions, probably declared in an anonymous namespace.

src/Featurizers/Components/VectorNormsEstimator.h
```diff
@@ -0,0 +1,192 @@
...
+namespace Microsoft {
+namespace Featurizer {
+namespace Featurizers {
+namespace Components {
+
+static constexpr char const * const         VectorNormsEstimatorName("VectorNormsEstimator");
+
+
+template <typename T>
+struct IsIteratorPair {
```
davidbrownellWork: >IsIteratorPair [](start = 7, length = 14)

Document this.

src/Featurizers/Components/VectorNormsEstimator.h
```diff
@@ -0,0 +1,192 @@
...
+    static constexpr bool const value = true;
+};
+template <typename T>
+struct IsIteratorPair<std::pair<T, T>> {
+    static constexpr bool const value = true;
+};
+
+/////////////////////////////////////////////////////////////////////////
+///  \class         VectorNormsAnnotationData
+///  \brief         An annotation class which contains the norms of a matrix
```
davidbrownellWork: >An annotation class which contains the norms of a matrix [](start = 20, length = 56)

Nit: consider change this description to "An annotation class which contains the norms for all rows of a matrix."

(This was confusing to me at first, so maybe the clarification will be helpful to others)

src/Featurizers/Components/VectorNormsEstimator.h
```diff
@@ -17,7 +17,9 @@ namespace Components {
 
 static constexpr char const * const         VectorNormsEstimatorName("VectorNormsEstimator");
 
-
+// vectors as input to VectorNormEstimator and NormalizeFeaturizer
+// are represented as a pair or a tuple of iterators std::pair<begin, end>
+// this struct is created for asserting a type to be a pair or a tuple of iterators
```
davidbrownellWork: Use doxygen format

src/Featurizers/NormalizeFeaturizer.h
```diff
@@ -181,55 +199,55 @@ class NormalizeEstimator :
...
+    ~NormalizeEstimatorImpl(void) override = default;
 
-    FEATURIZER_MOVE_CONSTRUCTOR_ONLY(NormalizeEstimator);
+    FEATURIZER_MOVE_CONSTRUCTOR_ONLY(NormalizeEstimatorImpl);
 };
 
 /////////////////////////////////////////////////////////////////////////
 ///  \typedef       L1NormalizeEstimator
-///  \brief         Used to differentiate different NormalizeEstimator depends on norms
+///  \brief         Used to differentiate different NormalizeEstimatorImpl depends on norms
```
davidbrownellWork: Update this comment so that it captures information from the user's perspective.

src/Featurizers/NormalizeFeaturizer.h
```diff
@@ -158,7 +176,7 @@ class NormalizeEstimatorImpl : public TransformerEstimator<IteratorRangeT, std::
 } // namespace Details
 
 /////////////////////////////////////////////////////////////////////////
-///  \class         NormalizeEstimator
+///  \class         NormalizeEstimatorImpl
```
davidbrownellWork: NormaizeEstimatorBase? and then change the Implx name back to Impl.

src/Tools/CodeGenerator/Featurizers.yaml
```diff
@@ -787,15 +787,42 @@ featurizers:
     estimator_name: SparseNormalizerEstimator
     num_output_columns: 1
     description: |-
-      Computes the L1 and/or L2 norm for a provided data set.
+      Computes the L1 or L2 or max norm for a provided data set and normalize every row so that
```
davidbrownellWork: We will need to introduce 3 variations of this estimator: L1, L2, and Max.

PR #73, opened 2020-01-17
Add Numericalize featurizer

yuslepukhin:
Add impl, add test. This is like LabelEncoder but better.


src/Featurizers/NumericalizeFeaturizer.h
```diff
@@ -0,0 +1,276 @@
+// ----------------------------------------------------------------------
```
davidbrownellWork: This change looks good. After this goes in, can you take a stab at a PR that exposes this functionality in the SharedLibrary. This will involve:

1) Running the code generator to produce the SharedLibrary files (.h, .cpp) and SharedLibraryTest files (.h)
2) Writing a simple integration test that exercises the functionality generated in the SharedLibraryTest file

PR #76, opened 2020-01-17
PCAFeaturizer

wangyems:


src/Featurizers/UnitTests/PCAFeaturizer_UnitTests.cpp
```diff
@@ -0,0 +1,138 @@
+// ----------------------------------------------------------------------
```
davidbrownellWork: Can you add a test that validates the exception with the string "fit_impl() should not be called move than once in PCAFeaturizer" is thrown when expected? While you are doing this, can you add the same test for TruncatedSVDWrapperFeaturizer as well.

wangyems: sure

PR #78, opened 2020-01-18
MedianImputerFeaturizer

davidbrownellWork:


src/Featurizers/MedianImputerFeaturizer.h
```diff
@@ -0,0 +1,190 @@
...
+///
+template <typename InputT, typename TransformedT, bool InterpolateValuesV, size_t MaxNumTrainingItemsV>
+class MedianImputerEstimatorImpl : public TransformerEstimator<InputT, TransformedT> {
+public:
+    // ----------------------------------------------------------------------
+    // |
+    // |  Public Types
+    // |
+    // ----------------------------------------------------------------------
+    static_assert(Traits<InputT>::IsNullableType, "'InputT' must be a nullable type");
```
angryjinyan: should there also be a static_assert that nonstd::optional<TransformedT> is the same type as InputT?

davidbrownellWork: Yes, I'll add that.

davidbrownellWork: Not necessarily, as InputT might be something like `nonstd::optional<int>` or `float`, both of which are nullable types.

angryjinyan: From my understanding, an imputer replaces missing values to a specified value, either min, max, median or something else. In the case of min, max and median, does it make sense if somebody set InputT as `nonstd::optional<int>` and TransformedT as `float`?

davidbrownellWork: This is the scenario (input of `nonstd::optional<int>` and and output of `float`) that requires the different types. Also, this only applies to the MedianImputerFeaturizer.

If the number of inputs is odd, the middle value is selected as the Median. If the number of inputs is even, the media value is the average of the two. This is why the output type is float for int inputs.

Not all types support average, which is why the Estimator supports the `InterpolateValuesV` template parameter.

PR #81, opened 2020-01-21
Strings.h

wangyems:
A simple String library for StringDecorator and StringParsing in DocumentStatisticsEstimator


src/Strings.h
```diff
@@ -0,0 +1,388 @@
...
+    Details::ParseNgramCharwb(input, isDelimiter, ngramRangeMin, ngramRangeMax, callback);
+}
+
+template <
+    typename IteratorT,
+    typename IsDelimiterT
+>
+inline void ParseNgramCharwbCopy(std::string const &input,
+                                 IsDelimiterT const &isDelimiter,
+                                 size_t const &ngramRangeMin,
```
yuslepukhin: >size_t const &ngramRangeMin, [](start = 33, length = 28)

This is a primitive type, pass by value

src/Strings.h
```diff
@@ -0,0 +1,388 @@
...
+                             std::function<void (IteratorT&, IteratorT&)> const &callback) {
+
+    Details::ParseNgramCharwb(input, isDelimiter, ngramRangeMin, ngramRangeMax, callback);
+}
+
+template <
+    typename IteratorT,
+    typename IsDelimiterT
+>
+inline void ParseNgramCharwbCopy(std::string const &input,
```
yuslepukhin: >inline void ParseNgramChar [](start = 0, length = 26)

Public utility, please, provide Docs

src/Strings.h
```diff
@@ -0,0 +1,388 @@
...
+    //copy string to trim
+    std::string inputCopy(input);
+    Details::ParseNgramChar(inputCopy, ngramRangeMin, ngramRangeMax, callback);
+}
+
+template <
+    typename IteratorT,
+    typename IsDelimiterT
+>
+inline void ParseNgramCharwb(std::string &input,
```
yuslepukhin: >inline void ParseNgramCha [](start = 0, length = 25)

Docs

src/Strings.h
```diff
@@ -0,0 +1,388 @@
...
+                             IsDelimiterT const &isDelimiter,
+                             size_t const &ngramRangeMin,
```
yuslepukhin: >size_t [](start = 29, length = 6)

Pirmitive type, pass by value

src/Strings.h
```diff
@@ -0,0 +1,388 @@
...
+                           size_t const &ngramRangeMin,
+                           size_t const &ngramRangeMax,
+                           std::function<void (IteratorT&, IteratorT&)> const &callback) {
+
+    Details::ParseNgramChar(input, ngramRangeMin, ngramRangeMax, callback);
+}
+
+template <typename IteratorT>
+inline void ParseNgramCharCopy(std::string const &input,
+                               size_t const &ngramRangeMin,
```
yuslepukhin: >size_t [](start = 31, length = 6)

Ditto

src/Strings.h
```diff
@@ -0,0 +1,388 @@
...
+                               size_t const &ngramRangeMax,
+                               std::function<void (IteratorT&, IteratorT&)> const &callback) {
```
yuslepukhin: >IteratorT&, IteratorT& [](start = 51, length = 22)

This worries me. Binds to a non-const reference. I need an explanation for this. Usually, iterators are passed by value so the originals are not modified. Especially, the end iterator, why it is a non-const ref?

src/Strings.h
```diff
@@ -0,0 +1,388 @@
...
+
+    //copy string to trim
+    std::string inputCopy(input);
+    Details::ParseNgramWord(inputCopy, isDelimiter, ngramRangeMin, ngramRangeMax, callback);
+}
+
+template <typename IteratorT>
+inline void ParseNgramChar(std::string &input,
+                           size_t const &ngramRangeMin,
+                           size_t const &ngramRangeMax,
```
yuslepukhin: > size_t [](start = 26, length = 7)

Ditto

src/Strings.h
```diff
@@ -0,0 +1,388 @@
...
+                               size_t const &ngramRangeMax,
+                               std::function<void (IteratorT&, IteratorT&)> const &callback) {
+
+    //copy string to trim
+    std::string inputCopy(input);
+    Details::ParseNgramWord(inputCopy, isDelimiter, ngramRangeMin, ngramRangeMax, callback);
+}
+
+template <typename IteratorT>
+inline void ParseNgramChar(std::string &input,
```
yuslepukhin: >inline void [](start = 0, length = 11)

Docs

src/Strings.h
```diff
@@ -0,0 +1,388 @@
...
+                           size_t const &ngramRangeMin,
+                           size_t const &ngramRangeMax,
+                           std::function<void (IteratorT&, IteratorT&)> const &callback) {
```
yuslepukhin: >IteratorT&, IteratorT& [](start = 47, length = 22)

Ditto

src/Strings.h
```diff
@@ -0,0 +1,388 @@
...
+    typename IteratorT,
+    typename PredicateT
+>
+inline std::string TrimRight(IteratorT begin,
+                             IteratorT end,
+                             PredicateT isPredicate) {
+    IteratorT offset = begin;
+    IteratorT iter = begin;
+    while (iter != end) {
+        if (*iter != isPredicate)
```
yuslepukhin: >*iter != isPredicate [](start = 12, length = 20)

Predicate is usually a function that evaluates the current value. Please, rewire according to the conventions.

yuslepukhin: >if (*iter != isPredicate) [](start = 8, length = 25)

This code is not clear

src/Strings.h
```diff
@@ -0,0 +1,388 @@
...
+inline std::string ToUpper(IteratorT begin,
+                           IteratorT end) {
+    std::string result(begin, end);
+
+#if (defined _MSC_VER)
+#   pragma warning(push)
+#   pragma warning(disable: 4244) // conversion from 'unsigned int' to 'unsigned char', possible loss of data
+#endif
+
+    std::transform(result.begin(), result.end(), result.begin(), [](char c) { return std::toupper(c); });
```
yuslepukhin: >return std::toupper(c); [](start = 78, length = 23)

This handles only ASCII characters. You need to handle UTF-8.

src/Strings.h
```diff
@@ -0,0 +1,388 @@
...
+inline std::string ToLower(IteratorT begin,
+                           IteratorT end) {
+    std::string result(begin, end);
+
+#if (defined _MSC_VER)
+#   pragma warning(push)
+#   pragma warning(disable: 4244) // conversion from 'unsigned int' to 'unsigned char', possible loss of data
+#endif
+
+    std::transform(result.begin(), result.end(), result.begin(), [](char c) { return std::tolower(c); });
```
yuslepukhin: >return std::tolower(c) [](start = 78, length = 22)

Same, handles only ASCII

src/Strings.h
```diff
@@ -0,0 +1,388 @@
...
+
+template <typename IteratorT>
+inline std::string ToLower(IteratorT begin,
+                           IteratorT end) {
+    std::string result(begin, end);
+
+#if (defined _MSC_VER)
+#   pragma warning(push)
+#   pragma warning(disable: 4244) // conversion from 'unsigned int' to 'unsigned char', possible loss of data
+#endif
```
yuslepukhin: A simple cast with a comment should do the job here.

src/Strings.h
```diff
@@ -0,0 +1,388 @@
...
+namespace Details {
+
+template<typename IteratorT>
+inline std::vector<IteratorT> IteratorVectorGenerator(IteratorT const & begin,
+                                                     IteratorT const & end) {
+    std::vector<IteratorT> wordIterVector;
+    wordIterVector.reserve(static_cast<unsigned int>(std::distance(begin, end) + 1));
+
+    for (IteratorT strIter = begin; strIter != end; ++strIter)
+        wordIterVector.emplace_back(std::move(strIter));
```
yuslepukhin: >strIter [](start = 46, length = 7)

If you move the iterator it becomes empty or invalid, you can no longer operate on it after the first iteration.

src/Strings.h
```diff
@@ -0,0 +1,388 @@
...
+    wordIterVector.emplace_back(std::move(end));
```
yuslepukhin: >std::move(end) [](start = 32, length = 14)

End is const, can not be moved.

src/Strings.h
```diff
@@ -0,0 +1,388 @@
...
+
+namespace Microsoft {
+namespace Featurizer {
+namespace Details {
+
+template<typename IteratorT>
+inline std::vector<IteratorT> IteratorVectorGenerator(IteratorT const & begin,
+                                                     IteratorT const & end) {
+    std::vector<IteratorT> wordIterVector;
+    wordIterVector.reserve(static_cast<unsigned int>(std::distance(begin, end) + 1));
```
yuslepukhin: > wordIterVector.reserve(static_cast<unsigned int>(std::distance(begin, end) + 1)); [](start = 3, length = 82)

Expensive and unnecessary. reserve only make sense if the length is known in advance. In this case, re-allocation will amortize nicely.

src/Strings.h
```diff
@@ -0,0 +1,388 @@
...
+    return wordIterVector;
+}
+
+template <
+    typename IteratorT,
+    typename IsDelimiterT
+>
+inline void Parse(IteratorT const &begin,
+                  IteratorT const &end,
+                  IsDelimiterT const &isDelimiter,
```
yuslepukhin: >IsDelimiterT const &isDelimiter, [](start = 18, length = 32)

Can you make it a predicate?

src/Strings.h
```diff
@@ -0,0 +1,388 @@
...
+                  std::function<void (IteratorT&, IteratorT&)> const &callback) {
```
yuslepukhin: >IteratorT&, IteratorT& [](start = 38, length = 22)

Need an explanation or a doc for these callbacks

src/Strings.h
```diff
@@ -0,0 +1,388 @@
...
+
+    IteratorT left = begin;
+    IteratorT right = left;
+    while (left != end && *left == isDelimiter)
+        ++left;
+
+    right = left;
+    while (right != end) {
+        if (*right == isDelimiter) {
+            callback(left, right);
```
yuslepukhin: >callback(left, right); [](start = 12, length = 22)

According to the signature, the callback will modify the iterators but it is not a legal thing to do here since they are used for iteration in this function.

src/Strings.h
```diff
@@ -0,0 +1,388 @@
...
+                       std::function<void (IteratorT&, IteratorT&)> const &callback) {
+
+    std::sregex_iterator iter(begin, end, regexToken);
+    std::sregex_iterator iterEnd;
+
+    while (iter != iterEnd) {
+        std::string const iterStringConst = iter->str();
+        IteratorT beginRef = iterStringConst.begin();
+        IteratorT endRef = iterStringConst.end();
+        callback(beginRef, endRef);
```
yuslepukhin: >callback(beginRef, endRef); [](start = 8, length = 27)

Ditto

src/Strings.h
```diff
@@ -0,0 +1,388 @@
...
+        ++iter;
+    }
+}
+
+template <typename IteratorT>
+inline std::string TrimAndReplace(IteratorT begin,
+                                  IteratorT end) {
+    //replace all punctuations with spaces
+    std::transform(begin, end, begin, [](char c) {
```
yuslepukhin: >begin [](start = 31, length = 5)

This assumes two things: 1) The iterator is not const 2) the input iterator can be used as an output iterator. And this is all about strings. Is it possible to use standard string functions and standard algos?

src/Strings.h
```diff
@@ -0,0 +1,388 @@
...
+
+    std::string result(begin, end);
+    //remove duplicate space
+    result.erase(
+        std::unique(
+            result.begin(),
+            result.end(),
+            [=](char lhs, char rhs){
+                return (lhs == rhs) && (lhs == ' ');
+            }
```
yuslepukhin: This version of unique is only available in C++20.

src/Strings.h
```diff
@@ -0,0 +1,388 @@
...
+    std::string trimedString(TrimAndReplace(input.begin(), input.end()));
+
+    //wordIterPairVector is used to store the begin and end iterator of words in input
+    std::vector<std::pair<IteratorT, IteratorT>> wordIterPairVector;
+    Details::Parse<IteratorT, IsDelimiterT>(
+        trimedString.begin(),
+        trimedString.end(),
+        isDelimiter,
+        [&wordIterPairVector] (IteratorT &iterBegin, IteratorT &iterEnd) {
+            wordIterPairVector.emplace_back(std::make_pair(iterBegin, iterEnd));
```
yuslepukhin: >emplace_back [](start = 31, length = 12)

This syntax makes no improvement over push_back. On the other hand, emplace_back(iterBegin, iterEnd) makes it more efficient.

src/Strings.h
```diff
@@ -0,0 +1,388 @@
...
+        result.end()
+    );
+
+    return result;
+}
+
+template <typename IteratorT>
+inline void ParseNgramCharHelper(IteratorT const &begin,
+                                 IteratorT const &end,
+                                 size_t const &ngramRangeMin,
```
yuslepukhin: >size_t const &ngramRangeMin, [](start = 33, length = 28)

Primitive type

yuslepukhin:
The functions has no documentation. A clear documentation would help to produce a better implementation and would simplify review process.

src/Strings.h
```diff
@@ -0,0 +1,388 @@
...
+        std::string const iterStringConst = iter->str();
+        IteratorT beginRef = iterStringConst.begin();
+        IteratorT endRef = iterStringConst.end();
+        callback(beginRef, endRef);
+        ++iter;
+    }
+}
+
+template <typename IteratorT>
+inline std::string TrimAndReplace(IteratorT begin,
```
yuslepukhin: >TrimAndReplace [](start = 19, length = 14)

The name of the function does not match its functionality. It is not trimming.

src/Strings.h
```diff
@@ -0,0 +1,388 @@
...
+inline void ParseNgramCharHelper(IteratorT const &begin,
+                                 IteratorT const &end,
+                                 size_t const &ngramRangeMin,
+                                 size_t const &ngramRangeMax,
+                                 std::function<void (IteratorT&, IteratorT&)> const &callback) {
+
+    std::vector<IteratorT> wordIterVector(IteratorVectorGenerator(begin, end));
+
+    for (size_t strIterOffset = 0; strIterOffset < wordIterVector.size(); ++strIterOffset)  {
+        for (size_t ngramRangeVal = ngramRangeMin; ngramRangeVal <= ngramRangeMax; ++ngramRangeVal) {
```
yuslepukhin: >ngramRangeMax [](start = 68, length = 13)

Due to a lack of documentation it is not clear if the range is inclusive or not.

src/Strings.h
wangyems: Yes this version only support ASCII characters, I'll add a todo here and will resolve in a future PR

src/Strings.h
```diff
@@ -0,0 +1,388 @@
...
+            if (strIterOffset + ngramRangeVal < wordIterVector.size())
+                callback(wordIterVector[strIterOffset], wordIterVector[strIterOffset + ngramRangeVal]);
```
yuslepukhin: >strIterOffset + ngramRangeVal [](start = 71, length = 29)

Should we assert that the resulting iterator does not point beyond the vector? Operator[] does not check bounds.

src/Strings.h
```diff
@@ -0,0 +1,388 @@
...
+                                 IteratorT const &end,
+                                 size_t const &ngramRangeMin,
+                                 size_t const &ngramRangeMax,
+                                 std::function<void (IteratorT&, IteratorT&)> const &callback) {
+
+    std::vector<IteratorT> wordIterVector(IteratorVectorGenerator(begin, end));
+
+    for (size_t strIterOffset = 0; strIterOffset < wordIterVector.size(); ++strIterOffset)  {
+        for (size_t ngramRangeVal = ngramRangeMin; ngramRangeVal <= ngramRangeMax; ++ngramRangeVal) {
+            if (strIterOffset + ngramRangeVal < wordIterVector.size())
```
yuslepukhin: >if (strIterOffset + ngramRangeVal < wordIterVector.size()) [](start = 12, length = 58)

Should the loop terminate if this is not true? Or even both loops?

src/Strings.h
wangyems: fixed

src/Strings.h
wangyems: the callbacks are functions to do the string decoration(casing and trimming), as from a previous talk with David that we apply split(parse, ngram...) function before decorate functions

src/Strings.h
wangyems: sure

src/Strings.h
wangyems: the callback functions are used for string decorator so that modification can apply to original string. I'll add comments then

src/Strings.h
wangyems: changed it to a function

src/Strings.h
wangyems: it is inclusive. I'll add comments

src/Strings.h
wangyems: sure

PR #86, opened 2020-01-22
SharedLibrary export of multiple featurizers

davidbrownellWork:
Added SharedLibrary support for the following featurizers:

- NumericalzeFeaturizer
- MinMaxImputerFeaturizer
- MeanImputerFeaturizer
- MedianImputerFeaturizer
- ModeImputerFeaturizer


src/Tools/CodeGenerator/Featurizers.yaml
```diff
@@ -393,6 +366,137 @@ featurizers:
       - input_type: TInputDouble
         output_type: double
 
+  # ----------------------------------------------------------------------
+  - name: MeanImputerFeaturizer
```
michaelgsharp: Just curious, why are we switching from one `Imputer` that can do all of these to individual ones that can only do one thing?

davidbrownellWork: I had originally wanted to do it this way, but when I started implementing them I realized that there were different computational steps required to calculate the value to impute (for example, we don't need to calculate the median value (which is relatively expensive) when we only need the mean for imputation). I ended up organizing them by functionality rather than introducing runtime overhead by putting them all in one.

michaelgsharp: Makes sense. It will require more interop code this way, but if its better for performance than its worth it.

PR #91, opened 2020-01-23
DocumentStatisticsEstimator

wangyems:


src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -17,31 +21,6 @@ namespace Components {
...
-            right = left;
-        } else {
-            ++right;
-        }
-    }
-    if (left != right)
-        callback(left, right);
-}
-
 struct IterRangeComp {
```
davidbrownellWork: >struct IterRangeComp [](start = 0, length = 20)

This appears to be an implementation detail - move its definition to the Implementation section of the code.

davidbrownellWork:
>             if (s2 == e2 || *s2 < *s1) {

This is comparing the iterators - do you want to compare the values instead?

---
Refers to: src/Featurizers/Components/DocumentStatisticsEstimator.h:40 in fefce89. [](commit_id = fefce89d8bf8bd92fbdb8709792cbf8265fabb91, deletion_comment = False)

davidbrownellWork:
>             if (s2 == e2 || *s2 < *s1) {

I believe that the code above ensures that s2 will never equal e2 - change this to an assertion.

---
Refers to: src/Featurizers/Components/DocumentStatisticsEstimator.h:40 in fefce89. [](commit_id = fefce89d8bf8bd92fbdb8709792cbf8265fabb91, deletion_comment = False)

davidbrownellWork:
>         return (s2 != e2);

I don't think that we will ever hit this code, will we?

---
Refers to: src/Featurizers/Components/DocumentStatisticsEstimator.h:48 in fefce89. [](commit_id = fefce89d8bf8bd92fbdb8709792cbf8265fabb91, deletion_comment = False)

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -74,6 +53,21 @@ struct IterRangeComp {
 
 static constexpr char const * const         DocumentStatisticsEstimatorName("DocumentStatisticsEstimator");
 
+/////////////////////////////////////////////////////////////////////////
+///  \struct        FrequencyAndIndexStruct
+///  \brief         This struct is a combination of values of FrequencyMap and
+///                 IndexMap for the purpose of keys equivelence of two maps
+///
+struct FrequencyAndIndexStruct {
```
davidbrownellWork: >FrequencyAndIndexStruct [](start = 7, length = 23)

Nit: remove the Struct suffix.

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -74,6 +53,21 @@ struct IterRangeComp {
...
+    std::uint32_t const TermFrequency;                            // Words and the number of documents that it appears in
+    std::uint32_t const Index;
+
+    //When generating the DocumentStatisticsAnnotationData, it will call FrequencyAndIndexStruct copy constructor
+    //FEATURIZER_MOVE_CONSTRUCTOR_ONLY(FrequencyAndIndexStruct);
```
davidbrownellWork: >FEATURIZER_MOVE_CONSTRUCTOR_ONLY(FrequencyAndIndexStruct); [](start = 6, length = 58)

The code should not be copyable - let's figure out what is going on.

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -128,8 +119,10 @@ class DocumentStatisticsTrainingOnlyPolicy {
     //       rather than asci characters.
     using InputType                         = std::string;
     using InputTypeConstIterator            = std::string::const_iterator;
-    using StringDecorator                   = std::function<std::string (InputTypeConstIterator begin, InputTypeConstIterator end)>;
-    using IndexMap                          = DocumentStatisticsAnnotationData::IndexMap;
+    using StringDecorator                   = std::function<std::string (std::string)>;
+    using IndexMap                          = std::unordered_map<std::string, std::uint32_t>;
+    using FrequencyMap                      = IndexMap;
+    using RegexT                            = std::regex;
```
davidbrownellWork: >using RegexT                            = std::regex; [](start = 4, length = 53)

It doesn't look like this is being used.

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -138,13 +131,28 @@ class DocumentStatisticsTrainingOnlyPolicy {
     // ----------------------------------------------------------------------
     static constexpr char const * const     NameValue = DocumentStatisticsEstimatorName;
 
+    enum AnalyzerMethod {
```
davidbrownellWork: >enum AnalyzerMethod [](start = 4, length = 19)

`enum class AnalyzerMethod : unsigned char {`

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -138,13 +131,28 @@ class DocumentStatisticsTrainingOnlyPolicy {
...
+        WORD = 0,
```
davidbrownellWork: >WORD [](start = 8, length = 4)

Make these values PascalCase

davidbrownellWork: >0 [](start = 15, length = 1)

The first value should be 1 to avoid potential problems with languages that default initialize to 0.

src/Featurizers/Components/DocumentStatisticsEstimator.h
davidbrownellWork: >AnalyzerMethod [](start = 9, length = 14)

This should be represented as a custom_enum in Featurizers.yaml

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -194,8 +210,40 @@ class DocumentStatisticsTrainingOnlyPolicy {
...
+    // ----------------------------------------------------------------------
+    // |
+    // |  Public Types
+    // |
+    // ----------------------------------------------------------------------
+    using BaseType                          = TrainingOnlyEstimatorImpl<Details::DocumentStatisticsTrainingOnlyPolicy, MaxNumTrainingItemsV>;
+    using StringDecorator                   = Details::DocumentStatisticsTrainingOnlyPolicy::StringDecorator;
+    using IndexMap                          = Details::DocumentStatisticsTrainingOnlyPolicy::IndexMap;
+    using AnalyzerMethod                    = Details::DocumentStatisticsTrainingOnlyPolicy::AnalyzerMethod;
+    using RegexT                            = Details::DocumentStatisticsTrainingOnlyPolicy::RegexT;
```
davidbrownellWork: >using RegexT                            = Details::DocumentStatisticsTrainingOnlyPolicy::RegexT; [](start = 4, length = 96)

It doesn't look like this is used publicly

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -208,27 +256,32 @@ using DocumentStatisticsEstimator           = Components::TrainingOnlyEstimatorI
...
 // ----------------------------------------------------------------------
-inline DocumentStatisticsAnnotationData::DocumentStatisticsAnnotationData(FrequencyMap termFreq, IndexMap termIndex, std::uint32_t totalNumDocuments) :
-    TermFrequency(
-        std::move(
-            [&termFreq](void) -> FrequencyMap & {
-                if(termFreq.empty())
-                    throw std::invalid_argument("termFreq");
+FrequencyAndIndexStruct::FrequencyAndIndexStruct(std::uint32_t termFrequency, std::uint32_t index) :
+    //validate not necessary in this assignment
+    TermFrequency(std::move(termFrequency)),
```
davidbrownellWork: >TermFrequency(std::move(termFrequency)), [](start = 4, length = 40)

Is it valid for a termFrequency to be 0?

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -208,27 +256,32 @@ using DocumentStatisticsEstimator           = Components::TrainingOnlyEstimatorI
...
+bool FrequencyAndIndexStruct::operator==(FrequencyAndIndexStruct const &other) const {
+    return (TermFrequency == other.TermFrequency) && (Index == other.Index);
+}
+
+// ----------------------------------------------------------------------
+// |
+// |  DocumentStatisticsAnnotationData
+// |
+// ----------------------------------------------------------------------
+inline DocumentStatisticsAnnotationData::DocumentStatisticsAnnotationData(FrequencyAndIndexMap termFrequencyAndIndex, std::uint32_t totalNumDocuments) :
```
davidbrownellWork: >FrequencyAndIndexMap termFrequencyAndIndex, std::uint32_t totalNumDocuments [](start = 74, length = 75)

Is there a relationship between the values in `termFrequencyAndIndex` and `totalNumDocuments`? For example, should there ever be a `termFrequency` value that is > `totalNumDocuments`?

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -246,31 +299,246 @@ inline DocumentStatisticsAnnotationData::DocumentStatisticsAnnotationData(Freque
...
+    size_t colIndex,
+    StringDecorator decorator,
+    AnalyzerMethod analyzer,
+    std::string regexToken,
+    nonstd::optional<IndexMap> existingVocabulary,
+    nonstd::optional<std::uint32_t> topKTerms,
+    std::float_t minDf,
+    std::float_t maxDf,
+    std::uint32_t ngramRangeMin,
+    std::uint32_t ngramRangeMax
```
davidbrownellWork: Validate these values

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -246,31 +299,246 @@ inline DocumentStatisticsAnnotationData::DocumentStatisticsAnnotationData(Freque
...
+) :
+    _stringDecoratorFunc(std::move(decorator)),
```
davidbrownellWork: >std::move(decorator) [](start = 25, length = 20)

Validate decorator

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -246,31 +299,246 @@ inline DocumentStatisticsAnnotationData::DocumentStatisticsAnnotationData(Freque
...
+    _analyzer(std::move(analyzer)),
```
davidbrownellWork: >std::move(analyzer) [](start = 14, length = 19)

Validate

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -246,31 +299,246 @@ inline DocumentStatisticsAnnotationData::DocumentStatisticsAnnotationData(Freque
...
+    _regexToken(std::move(regexToken)),
```
davidbrownellWork: >std::move(regexToken) [](start = 16, length = 21)

Validate

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -285,7 +553,7 @@ inline void Details::DocumentStatisticsTrainingOnlyPolicy::fit(InputType const &
 
         fit_impl<InputTypeConstIteratorRangeSet>(
             input,
-            [](InputTypeConstIterator begin, InputTypeConstIterator end) {
+            [](InputTypeConstIterator & begin, InputTypeConstIterator & end) {
```
davidbrownellWork: >[](InputTypeConstIterator & begin, InputTypeConstIterator & end) [](start = 12, length = 64)

`[](InputTypeConstIterator begin, InputTypeConstIterator end)`

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -307,12 +588,56 @@ template <typename SetT, typename CreateKeyFuncT, typename KeyToStringFuncT>
 void Details::DocumentStatisticsTrainingOnlyPolicy::fit_impl(InputType const &input, CreateKeyFuncT const &createKeyFunc, KeyToStringFuncT const &keyToStringFunc) {
     SetT                                    documents;
 
-    split_temp(
-        input,
-        [&createKeyFunc, &documents](InputTypeConstIterator begin, InputTypeConstIterator end) {
-            documents.insert(createKeyFunc(begin, end));
+    if (_analyzer == AnalyzerMethod::WORD) {
```
davidbrownellWork: >if (_analyzer == AnalyzerMethod::WORD) { [](start = 4, length = 40)

The behavior captured in these if statements will be the same for the lifetime of the class. Consider creating a `std::function` set during construction that can be called here directly.

davidbrownellWork:

I'm not sure that these tests coversall scenarios. Let's finalize the header file and then review this code.

---
Refers to: src/Featurizers/Components/UnitTests/DocumentStatisticsEstimator_UnitTest.cpp:1 in fefce89. [](commit_id = fefce89d8bf8bd92fbdb8709792cbf8265fabb91, deletion_comment = False)

src/Featurizers/Components/DocumentStatisticsEstimator.h
wangyems: it is used in fit_impl

src/Featurizers/Components/DocumentStatisticsEstimator.h
wangyems: have put it privately

src/Featurizers/Components/DocumentStatisticsEstimator.h
wangyems: I don't think we have to because if(!decorator) and if(decorator) are two scenarios in fit_impl

src/Featurizers/Components/DocumentStatisticsEstimator.h
wangyems: the same with decorator

src/Featurizers/Components/DocumentStatisticsEstimator.h
wangyems: it won't happen, will add validation there

src/Featurizers/Components/DocumentStatisticsEstimator.h
wangyems: no relationship

src/Featurizers/Components/DocumentStatisticsEstimator.h
wangyems: solved

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -86,27 +49,28 @@ class DocumentStatisticsAnnotationData {
...
     std::uint32_t const                     TotalNumDocuments;
 
     // ----------------------------------------------------------------------
     // |
     // |  Public Methods
     // |
     // ----------------------------------------------------------------------
-    DocumentStatisticsAnnotationData(FrequencyMap termFreq, IndexMap termIndex, std::uint32_t totalNumDocuments);
-    ~DocumentStatisticsAnnotationData(void) = default;
+    DocumentStatisticsAnnotationData(DocumentStatisticsAnnotationData &&other) :
```
davidbrownellWork: Separate declaration and definition; also move this standard constructor above this move constructor.

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -86,27 +49,28 @@ class DocumentStatisticsAnnotationData {
...
+        TermFrequencyAndIndex(std::move(const_cast<FrequencyAndIndexMap &>(other.TermFrequencyAndIndex))),
+        TotalNumDocuments(std::move(other.TotalNumDocuments)) {
+    }
 
-    FEATURIZER_MOVE_CONSTRUCTOR_ONLY(DocumentStatisticsAnnotationData);
+    DocumentStatisticsAnnotationData(FrequencyAndIndexMap termFrequencyAndIndex, std::uint32_t totalNumDocuments);
```
davidbrownellWork: Does this class need a move assignment operator as well?

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -155,14 +135,23 @@ class DocumentStatisticsTrainingOnlyPolicy {
     // |  Private Types
     // |
     // ----------------------------------------------------------------------
-    using FrequencyMap                      = typename DocumentStatisticsAnnotationData::FrequencyMap;
-
+    using FrequencyAndIndexMap              = DocumentStatisticsAnnotationData::FrequencyAndIndexMap;
+    using RegexT                            = std::regex;
```
davidbrownellWork: This type seems to only be used in fit_impl - can it be defined there?

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -246,31 +279,276 @@ inline DocumentStatisticsAnnotationData::DocumentStatisticsAnnotationData(Freque
...
+        }
+    }
+
+    //trim by topKTerms
+    if (!topKTerms.has_value() || termFrequency.empty() || *topKTerms >= termFrequency.size())
+        return termFrequency;
+
+    std::uint32_t const & topKTermsValue(*topKTerms);
+
+    std::tuple<std::uint32_t, long long> targetPack(
```
davidbrownellWork: This code could use some comments. I think that you are calculating the minimum value for the number of occurrences that should be kept and also the total number of items that should be kept?

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -246,31 +279,276 @@ inline DocumentStatisticsAnnotationData::DocumentStatisticsAnnotationData(Freque
...
+
+    std::uint32_t targetValue = std::get<0>(targetPack);
+    long long remainingTargetValue = std::get<1>(targetPack);
+
+    FrequencyMap::iterator termFrequencyIter = termFrequency.begin();
+    while (termFrequencyIter != termFrequency.end()) {
+        if (termFrequencyIter->second < targetValue) {
+            termFrequencyIter = termFrequency.erase(termFrequencyIter);
+            continue;
+        } else if (termFrequencyIter->second == targetValue) {
```
davidbrownellWork: In the first if clause, you are continuing, however, in this one you are not. I'd suggest writing this loop so that you leverage continue if the iterator has been updated due to an erase and only increment the iterator in one place.

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -246,31 +279,276 @@ inline DocumentStatisticsAnnotationData::DocumentStatisticsAnnotationData(Freque
...
+
+    return termFrequency;
+}
+
+inline DocumentStatisticsAnnotationData::FrequencyAndIndexMap MergeTwoMapsWithSameKeys(Details::DocumentStatisticsTrainingOnlyPolicy::FrequencyMap termFrequency,
+                                                                                       Details::DocumentStatisticsTrainingOnlyPolicy::IndexMap termIndex) {
+    DocumentStatisticsAnnotationData::FrequencyAndIndexMap termFrequencyAndIndex;
+    for (auto const & termFrequencyIter : termFrequency) {
+        //assert that termFrequencyIter.first is in termIndex
+        auto termIndexIter = termIndex.find(termFrequencyIter.first);
```
davidbrownellWork: Don't use auto here.

src/Featurizers/Components/DocumentStatisticsEstimator.h
davidbrownellWork: Add comments or decorate the name to indicate that it is an optional param.

src/Featurizers/Components/DocumentStatisticsEstimator.h
```diff
@@ -296,23 +574,84 @@ inline void Details::DocumentStatisticsTrainingOnlyPolicy::fit(InputType const &
...
 // ----------------------------------------------------------------------
 template <typename SetT, typename CreateKeyFuncT, typename KeyToStringFuncT>
 void Details::DocumentStatisticsTrainingOnlyPolicy::fit_impl(InputType const &input, CreateKeyFuncT const &createKeyFunc, KeyToStringFuncT const &keyToStringFunc) {
     SetT                                    documents;
 
-    split_temp(
-        input,
-        [&createKeyFunc, &documents](InputTypeConstIterator begin, InputTypeConstIterator end) {
-            documents.insert(createKeyFunc(begin, end));
+    if (_analyzer == AnalyzerMethod::Word) {
```
davidbrownellWork: I have the same comment as before - this logic can be invoked once at class construction time rather than during each call to fit.

src/Featurizers/Components/DocumentStatisticsEstimator.h
wangyems: it looks both AnnotationClass and This class do not need assignment operator

src/Featurizers/Components/DocumentStatisticsEstimator.h
wangyems: yes -I haven't marked this as resolved, will soon

src/Featurizers/Components/DocumentStatisticsEstimator.h
wangyems: sure

src/Featurizers/Components/DocumentStatisticsEstimator.h
wangyems: Comment it as optional parameter

src/Featurizers/Components/DocumentStatisticsEstimator.h
wangyems: fixed

src/Featurizers/Components/DocumentStatisticsEstimator.h
wangyems: fixed

src/Featurizers/Components/DocumentStatisticsEstimator.h
wangyems: commented

wangyems: Eg: assuming i have an vector [2,3,3,7,7,6,5,4,4,4,4,4] and the aim is to find top k values(if k = 6), after partial sorting we know the 6th largest value is 4(the targetValue), and there are one 4 ahead of it(remainingTargetValue)

src/Featurizers/Components/DocumentStatisticsEstimator.h
wangyems: there's place for no increment, I'm not sure how to use loop here. I'll remain the original one for now

src/Featurizers/Components/DocumentStatisticsEstimator.h
wangyems: will make an update tomorrow

PR #101, opened 2020-01-25
Matrix template args

davidbrownellWork:
Updated featurizers using Eigen::Matrix types so that different Matrix variations can be provided.

### DIFF
--- a/src/Featurizers/PCAFeaturizer.h
+++ b/src/Featurizers/PCAFeaturizer.h
@@ -17,22 +17,27 @@ namespace Featurizers {
 ///                 matrix for dimensionality reduction, also provides
 ///                 PCAComponents retriving
 ///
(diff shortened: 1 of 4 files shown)


src/Featurizers/PCAFeaturizer.h
```diff
@@ -43,16 +48,13 @@ class PCATransformer : public StandardTransformer<MatrixT, MatrixT> {
 
     void save(Archive &ar) const override;
 
-    std::ptrdiff_t getEigenVectorRowsNumber() const;
```
wangyems: Dmitri wants shape information to be exposed, so that I added these two funcs

davidbrownellWork: Looks like a merge problem - I will figure out what happened and update the PR.

PR #104, opened 2020-01-25
add fuzzy comparison in Struct.h

wangyems:


src/Featurizers/Structs.h
```diff
@@ -120,14 +123,28 @@ SparseVectorEncoding<T>::ValueEncoding::ValueEncoding(value_type value, std::uin
...
+
 #if (defined __clang__)
 #   pragma clang diagnostic push
 #   pragma clang diagnostic ignored "-Wfloat-equal"
 #endif
 
 template <typename T>
 bool SparseVectorEncoding<T>::ValueEncoding::operator==(ValueEncoding const &other) const {
-    return Value == other.Value
+    return static_cast<T>(abs(Value - other.Value)) < static_cast<T>(EPS)
```
davidbrownellWork: If you make this change here, we need to make similar changes to all equality operators. Can the Catch2 Approx functionality give you what you need for now?

See SharedLibrary_NumericalizerFeaturizer_IntegrationTest.cpp for an example.

wangyems: yeah I can use that to customize my test function. I'll close this PR

PR #107, opened 2020-01-25
Add unit test for PCA and SVD

angryjinyan:
Add one additional unit test with different training and inferencing data for PCAFeaturizer and SVDFeaturizer.

### DIFF
--- a/src/Featurizers/TruncatedSVDFeaturizer.h
+++ b/src/Featurizers/TruncatedSVDFeaturizer.h
@@ -72,6 +72,26 @@ private:
     }
 };
 
+template<typename MatrixType>
+inline void svd_flip(MatrixType& mat) {
+    using Scalar = typename std::decay<decltype(mat(0,0))>::type;
+
+    typedef typename MatrixType::Index Index;
+
+    for(Index col = 0; col < mat.cols(); ++col)
+    {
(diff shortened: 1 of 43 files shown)


src/Featurizers/TruncatedSVDFeaturizer.h
```diff
@@ -11,6 +11,27 @@ namespace Microsoft {
 namespace Featurizer {
 namespace Featurizers {
 
+template<typename MatrixType>
+inline MatrixType& svd_flip(MatrixType& mat) {
+    using Scalar = typename std::decay<decltype(mat(0,0))>::type;
```
davidbrownellWork: Nit: this can be written as `using Scalar = typename MatrixType::Scalar;`

src/Featurizers/TruncatedSVDFeaturizer.h
```diff
@@ -11,6 +11,27 @@ namespace Microsoft {
 namespace Featurizer {
 namespace Featurizers {
 
+template<typename MatrixType>
+inline MatrixType& svd_flip(MatrixType& mat) {
```
davidbrownellWork: We should add comments that explain what this function is doing and why it is necessary.

PR #110, opened 2020-01-27
fix strings.h 

wangyems:
1, fix issues of dangling iterator
2, write uniform wrappers for concise calling from DocumentStatisticEstimator
3, the vector allocation is hardcoded which should be avoid, an introduction of another intermediateValues or using a linkedlist will address the problem. However I am wondering if there has a better solution

### DIFF
--- a/src/Strings.h
+++ b/src/Strings.h
@@ -464,7 +464,7 @@ void ParseNgramWord(std::string &input,
                     size_t const ngramRangeMax,
                     std::function<void (IteratorT, IteratorT)> const &callback) {
 
-    intermediateValues.emplace_back(std::move(Details::ReplaceAndDeDuplicate<std::function<bool (char)>>(input)));
+    intermediateValues.emplace_back(Details::ReplaceAndDeDuplicate<std::function<bool (char)>>(input));
 
     //wordIterPairVector is used to store the begin and end iterator of words in input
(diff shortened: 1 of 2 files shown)


src/Strings.h
```diff
@@ -70,6 +70,7 @@ template <
     typename RegexT
 >
 void ParseRegex(std::string const &input,
+                std::vector<std::string> &intermediateValues,
```
davidbrownellWork: >std::vector<std::string> &intermediateValues, [](start = 16, length = 45)

Let's discuss this in person - this signature doesn't feel right to me.

src/Strings.h
```diff
@@ -472,14 +581,96 @@ template <
...
                           size_t const ngramRangeMax,
                           std::function<void (IteratorT, IteratorT)> const &callback) {
 
     //copy string to trim
     std::string inputCopy(input);
-    ParseNgramCharwb(inputCopy, predicate, ngramRangeMin, ngramRangeMax, callback);
+    ParseNgramCharwb(inputCopy, intermediateValues, predicate, ngramRangeMin, ngramRangeMax, callback);
+}
+
+//to uniform so that all parse functions have same number of signature
```
davidbrownellWork: This doesn't feel right either. It isn't clear to me why these changes are necessary.

PR #112, opened 2020-01-27
strings.h fix

wangyems:


src/Strings.h
```diff
@@ -178,6 +142,19 @@ void ParseNgramCharwbCopy(std::string const &input,
 // ----------------------------------------------------------------------
 namespace Details {
 
+template <typename UnaryPredicateT>
+std::string StringPadding(std::string input, UnaryPredicateT isPredicate) {
+    if (!isPredicate(input.at(0)) && !isPredicate(input.at(input.length() - 1))) {
```
davidbrownellWork: I'm guessing that these predicates are not expensive to call. However, they may be. Consider assigning both results to a bool value and then using those values in the conditions so that they are only invoked once.

src/Strings.h
```diff
@@ -390,97 +370,52 @@ void ParseNgramWord(std::string &input,
...
 
     if (ngramRangeMin < 1 || ngramRangeMin > ngramRangeMax )
         throw std::invalid_argument("ngramRangeMin and ngramRangeMax not valid");
 
-    for (size_t pairIdx = 0; pairIdx < wordIterPairVector.size(); ++pairIdx) {
-        //copy
-        std::string const paddedWord = " " + std::string(wordIterPairVector[pairIdx].first, wordIterPairVector[pairIdx].second) + " ";
-
-        Details::ParseNgramCharHelper<IteratorT>(paddedWord.begin(), paddedWord.end(), ngramRangeMin, ngramRangeMax, callback);
+    for (size_t pairIdx = 0; pairIdx < wordIterPairVector.size() - 1; ++pairIdx) {
```
davidbrownellWork: This code strikes me as a little bit odd, as it using using the first and second values from different iterators. Is this intended? If so, add a comment explaining what is happening.

wangyems: comments added

PR #119, opened 2020-01-29
Added functionality to return empty strings for null values

davidbrownellWork:


src/Featurizers/StringFeaturizer.h
```diff
@@ -88,9 +137,85 @@ class StringEstimator : public Components::InferenceOnlyEstimatorImpl<StringTran
...
+        return "";
+
+    return ToString(value);
+}
+
+template <typename T>
+// static
+std::string StringTransformer<T>::ToStringWithEmptyStringsForNullValuesImpl(T const &, std::false_type) {
+    throw std::runtime_error("This should never be called");
 }
```
michaelgsharp: If this won't ever be called them why do we need it? Is it to make the compiler happy by having both the `std::false_type` and the `std::true_type`?

davidbrownellWork: The compiler complains if both methods aren't defined, even though one of them will never be called. I tried to create the declaration (but not the definition), but the compiler seemed to want to see the definition as well.

src/Featurizers/StringFeaturizer.h
```diff
@@ -16,61 +16,110 @@ namespace Featurizers {
 ///  \brief         Converts input into strings.
 ///
 template <typename T>
-class StringTransformer : public Components::InferenceOnlyTransformerImpl<T, std::string> {
+class StringTransformer : public StandardTransformer<T, std::string> {
```
michaelgsharp: This is more of an academic question, but when I was starting this change I left it as an `InferenceOnlyTransofmrer` and just overrode the methods that needed changing. Just curious what the pros/cons are to that approach vs this one that changes it to a `StandardTransformer`?

src/Featurizers/StringFeaturizer.h
```diff
@@ -88,9 +137,85 @@ class StringEstimator : public Components::InferenceOnlyEstimatorImpl<StringTran
 // |  StringTransformer
 // |
 // ----------------------------------------------------------------------
+template <typename T>
+StringTransformer<T>::StringTransformer(bool useEmptyStringsForNullValues /*=false*/) :
+    _useEmptyStringsForNullValues(
+        std::move(
+            [&useEmptyStringsForNullValues](void) -> bool & {
```
michaelgsharp: Since we just talked about reference capture in lambdas, just want to make sure I understand right. this one is ok because the lambda is part of the class and so is the boolean value, is that correct?

src/Featurizers/StringFeaturizer.h
```diff
@@ -88,9 +137,85 @@ class StringEstimator : public Components::InferenceOnlyEstimatorImpl<StringTran
...
+                if(useEmptyStringsForNullValues && Traits<T>::IsNullableType == false)
+                    throw std::invalid_argument("empty strings cannot be used with types that are not nullable");
```
michaelgsharp: Is it better to throw an error here or just default it back to the normal `ToString` approach?

src/Featurizers/StringFeaturizer.h
```diff
@@ -88,9 +137,85 @@ class StringEstimator : public Components::InferenceOnlyEstimatorImpl<StringTran
...
+                throw std::runtime_error("Unsupported archive version");
+
+            return StringTransformer(Traits<bool>::deserialize(ar));
+        }()
+    ) {
+}
+
+template <typename T>
+bool StringTransformer<T>::operator==(StringTransformer const &other) const {
+    return _useEmptyStringsForNullValues == other._useEmptyStringsForNullValues;
```
michaelgsharp: I have never done an `==` with templated classes. Will it handle the type checking automatically? Or will it say that `StringTransformer<float> == StringTransformer<double>` if they both have `true` for `_useEmptyStringsForNullValues `?

src/Featurizers/StringFeaturizer.h
davidbrownellWork: The lambda is invoked immediately, so the lifetime of the referenced value is greater than the lifetime of the lambda.

src/Featurizers/StringFeaturizer.h
davidbrownellWork: I prefer that the programmer be intentional about the behavior that they want when opting into less common scenarios.

src/Featurizers/StringFeaturizer.h
davidbrownellWork: That would result in a compiler error (because the type `StringTransformer<float>` is not the same as the type `StringTransformer<double>`. If we wanted to implement cross-type equality, we could redefine this as `template <typename OtherT> bool operator==(StringTransformer<OtherT> const &other) const`.

PR #120, opened 2020-01-29
TfidfVectorizer

wangyems:


src/Featurizers/TfidfVectorizerFeaturizer.h
```diff
@@ -0,0 +1,646 @@
...
+                } else {
+                    idf = 1.0 + std::log((1 + TotalNumsDocuments) / (0.0 + DocumentFreq.at(word)));
+                }
+
+                //calculate tfidf (tfidf = tf * idf)
+                std::float_t tfidf = static_cast<std::float_t>(tf * idf);
+
+                //calculate normVal
+                if(Norm == NormMethod::L1) {
+                    normVal += tfidf;
```
angryjinyan: Since abs() is removed here, we should add an assert(tfidf >=0 ) and normVal(>= 0)

src/Featurizers/TfidfVectorizerFeaturizer.h
```diff
@@ -0,0 +1,646 @@
...
+                //temperarily put output in a vector for future normalization
+                results.emplace_back(std::make_tuple(labelIter->second, tfidf));
+            }
+        }
+        //normVal will never be 0 as long as results is not empty
+        assert(normVal != 0.0f);
+
+        // l2-norm calibration
+        if (Norm == NormMethod::L2)
+            normVal = sqrt(normVal);
```
angryjinyan: and maybe same assert here

src/Featurizers/TfidfVectorizerFeaturizer.h
wangyems: added

src/Featurizers/TfidfVectorizerFeaturizer.h
wangyems: I believe the assert(normVal != 0.0f) is enough as the calculation ensure that if (normVal != 0) then it will > 0

src/Featurizers/TfidfVectorizerFeaturizer.h
```diff
@@ -0,0 +1,646 @@
...
+
+        // ----------------------------------------------------------------------
+        using DocumentStatisticsAnnotationData          = Components::DocumentStatisticsAnnotationData;
+        using DocumentStatisticsEstimator               = Components::DocumentStatisticsEstimator<MaxNumTrainingItemsV>;
+
+        // ----------------------------------------------------------------------
+
+        DocumentStatisticsAnnotationData const &        data(DocumentStatisticsEstimator::get_annotation_data(BaseType::get_column_annotations(), _colIndex, Components::DocumentStatisticsEstimatorName));
+
+
```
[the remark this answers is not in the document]
wangyems: removed

PR #122, opened 2020-01-29
Fix Eigen:Map type mapping

yuslepukhin:


.gitignore
```diff
@@ -17,6 +17,7 @@ CmakeSettings.json
 
 **/.vscode/*
 **/build/*
+**/build_17/*
```
davidbrownellWork: Let's remove this.

src/SharedLibrary/GeneratedCode/SharedLibrary_PCAFeaturizer.cpp
```diff
@@ -36,7 +36,8 @@ FEATURIZER_LIBRARY_API bool PCAFeaturizer_float_CreateEstimator(/*out*/ PCAFeatu
         *ppErrorInfo = nullptr;
 
         // No validation
-        Microsoft::Featurizer::Featurizers::PCAEstimator<Eigen::Map<Eigen::MatrixX<float>>>* pEstimator = new Microsoft::Featurizer::Featurizers::PCAEstimator<Eigen::Map<Eigen::MatrixX<float>>>(std::make_shared<Microsoft::Featurizer::AnnotationMaps>(1), 0 );
+        Microsoft::Featurizer::Featurizers::PCAEstimator<Eigen::Map<const Eigen::MatrixX<float>>>* pEstimator =
```
davidbrownellWork: Please update the CodeGenerator used to generate these files rather than the files themselves. These changes will be lost the next time the code generator is run.

.gitignore
yuslepukhin: I need this

PR #131, opened 2020-02-03
Strings.h fix

wangyems:
1, add std header file


src/Strings.h
```diff
@@ -213,8 +216,8 @@ inline void ParseRegex(IteratorT const &begin,
                        RegexT const &regexToken,
                        std::function<void (IteratorT, IteratorT)> const &callback) {
 
-    std::sregex_iterator iter(begin, end, regexToken);
-    std::sregex_iterator iterEnd;
+    std::regex_iterator<IteratorT> iter(begin, end, regexToken);
```
davidbrownellWork: Please add tests that exercise this functionality.

wangyems: We already have that: line 62 in Strings_UnitTest.cpp

davidbrownellWork: I believe that that test doesn't expose the problem because it is using string iterators as the underlying type; we need to have another test that uses a different iterator type (char pointers for example).

yuslepukhin:
> inline std::string ToLower(std::string input);

Is it a good time to ask a question why many of them take strings by value instead of by reference?

---
Refers to: src/Strings.h:22 in 729c7f5. [](commit_id = 729c7f5c3ebc3905904cd0a9c6344ef967e01dae, deletion_comment = False)

src/Strings.h
wangyems: have undo this

src/Strings.h
```diff
@@ -13,22 +19,22 @@ namespace Strings {
 ///  \brief         lowercase string
 ///                 only support ASCII, will do UTF-8 in later PR
 ///
-inline std::string ToLower(std::string input);
+inline std::string ToLower(std::string const & input);
```
davidbrownellWork: With this change, we lose the ability to move into the parameter - the parameter should be by value (in the support of move) if it is going to change. @yuslepukhin, please comment if you know of a better way to active the same results.

yuslepukhin: Here is my 2 cents: 1) As a general rule we don't want input parms to be modified 2) This function does need to move anything, it needs a copy 3) functions that accept movable parameters should be clearly marked as std::string& which give it sematic meaning, compiler checks and IDE support. This way, you can see that the original var looses its state and avoids lots of bugs. It will also prompt the user to supply std::move(). This signature does not enforce it so it is easy not to do it and do a copy.
Bottom line: design a function with clear intent (which is extremely simple in this case). 
I do not have a strong opinion in this case. But I am expecting lots of extra copies bc I would never move anything to a function with a non-movable parameter.


---
In reply to: [374383490](https://github.com/microsoft/FeaturizersLibrary/pull/131#discussion_r374383490) [](ancestors = 374383490)

davidbrownellWork: `std::string ToLower(std::string && input);`

might be more explict according to https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#f15-prefer-simple-and-conventional-ways-of-passing-information.

Thoughts on pros/cons?

yuslepukhin: What you, I think, really want is a combination of const std::string& and std::string&&. The reason being for that is the implementation makes a copy of the string which it does not need to do. My proposal in this case is to lean towards const std::string&. Not to make copies, create an output result and output there.

---
In reply to: [374389257](https://github.com/microsoft/FeaturizersLibrary/pull/131#discussion_r374389257) [](ancestors = 374389257)

yuslepukhin: One more thing I forgot. While std::string&& does not force the implementation to move, it is expected to move, hence my preference for const ref.

---
In reply to: [374390769](https://github.com/microsoft/FeaturizersLibrary/pull/131#discussion_r374390769) [](ancestors = 374390769,374389257)

davidbrownellWork: Let's keep these as value parameters per https://web.archive.org/web/20140205194657/http://cpp-next.com/archive/2009/08/want-speed-pass-by-value/.

PR #132, opened 2020-02-03
Fix spelling error of scaler

angryjinyan:
Fix all ScalerFeaturizers spelling


src/Tools/CodeGenerator/Featurizers.yaml
```diff
@@ -479,8 +479,8 @@ featurizers:
         is_optional: false
 
   # ----------------------------------------------------------------------
-  - name: MaxAbsScalarFeaturizer
-    estimator_name: MaxAbsScalarEstimator
+  - name: MaxAbsScalerFeaturizer
```
davidbrownellWork: @michaelgsharp / @yuslepukhin: What is the impact of this change to ML.Net and ORT? Do we need to have a migration plan in place or is this a change that we can take immediately?

yuslepukhin: I do not see a need for a migration plan since there are no models given to customers (right?) However, some of the unit tests refer to estimators classes by name and that will need to be adjusted.

---
In reply to: [374429103](https://github.com/microsoft/FeaturizersLibrary/pull/132#discussion_r374429103) [](ancestors = 374429103)

PR #133, opened 2020-02-04
Refactor of CountVectorizer

wangyems:


wangyems:
The C Wrapper is not updated yet so it won't pass the test

src/Featurizers/CountVectorizerFeaturizer.h
```diff
@@ -175,7 +23,7 @@ template <size_t MaxNumTrainingItemsV=std::numeric_limits<size_t>::max()>
 class CountVectorizerEstimator :
     public Components::PipelineExecutionEstimatorImpl<
         Components::DocumentStatisticsEstimator<MaxNumTrainingItemsV>,
-        Details::CountVectorizerEstimatorImpl<MaxNumTrainingItemsV>
+        Details::TfidfVectorizerEstimatorImpl<MaxNumTrainingItemsV>
```
davidbrownellWork: >Details::TfidfVectorizerEstimatorImpl [](start = 8, length = 37)

We shouldn't be reaching into the Details namespace for this functionality. Can the code contain a member instance of TfidfVectorizerEstimator?

src/Featurizers/TfidfVectorizerFeaturizer.cpp
```diff
@@ -201,13 +201,16 @@ void TfidfVectorizerTransformer::execute_impl(typename BaseType::InputType const
             results.emplace_back(std::make_tuple(labelIter->second, tfidf));
         }
     }
-    //normVal will never be 0 as long as results is not empty
-    assert(normVal > 0.0f);
+    //normVal will be zero when the input is empty
+    assert(normVal >= 0.0f);
```
davidbrownellWork: >assert(normVal >= 0.0f); [](start = 4, length = 24)

I think that the logic is this:

the normValue will never be 0 if the NormMethod is L1 or L2. It will always be 0 if the NormMethod is None. 

Can you refactor this assert to make that explicit?

src/Featurizers/UnitTests/CountVectorizerFeaturizer_UnitTests.cpp
```diff
@@ -72,9 +56,10 @@ TEST_CASE("empty inferencing data - without binary without decorator") {
         ), "'values' is empty"
     );
 }
+
 TEST_CASE("string - without binary, without decorator, analyze word, maxdf = 1, mindf = 0, topk = null, empty vocabulary, ngram_min = 1, ngram_max = 1") {
     using InputType       = std::string;
-    using TransformedType = NS::Featurizers::SparseVectorEncoding<std::uint32_t>;
+    using TransformedType = NS::Featurizers::SparseVectorEncoding<std::float_t>;
```
davidbrownellWork: >float_t [](start = 71, length = 7)

Will CountVectorizer ever return a value that is not a whole number? I think that CountVectorizer should retain the existing signature; values should be converted from what is returned by the underlying implementation.

src/Tools/CodeGenerator/Featurizers.yaml
```diff
@@ -78,7 +78,7 @@ featurizers:
 
     type_mappings:
       - input_type: string
-        output_type: sparse_vector<uint32>
+        output_type: sparse_vector<float>
```
davidbrownellWork: >float [](start = 35, length = 5)

Per other comments, I think that this should remain uint32.

src/Featurizers/TfidfVectorizerFeaturizer.cpp
wangyems: I realized yesterday that when the input is empty, the normValue is 0

src/Featurizers/CountVectorizerFeaturizer.h
wangyems: I'll reach out to discuss this as I am not sure how to do it

PR #135, opened 2020-02-05
Introduced new featurizer metadata

davidbrownellWork:
- release_version (required)
- updated_version
- deprecated_version
- is_stateless_transformer


src/Tools/CodeGenerator/Featurizers.SimpleSchema
```diff
@@ -24,10 +24,21 @@
 <featurizers +>:
     [name string]
     [estimator_name string]
+
+    [release_version string]
+    [updated_version string ?]
+    [deprecated_version string ?]
```
michaelgsharp: Whats the difference between release_version and updated_version?

src/Tools/CodeGenerator/Featurizers.yaml
```diff
@@ -2,6 +2,10 @@ featurizers:
   # ----------------------------------------------------------------------
   - name: CatImputerFeaturizer
     estimator_name: CatImputerEstimator
+
+    release_version: 0.3.0
+    deprecated_version: 0.3.6
```
michaelgsharp: Are these flipped? If I am reading right it seems the deprecated_version is newer then the release_version.

src/Tools/CodeGenerator/Featurizers.SimpleSchema
davidbrownellWork: Release_version is when the featurizer was released, updated_version is the version when the featurizer was last updated after its release.

src/Tools/CodeGenerator/Featurizers.yaml
davidbrownellWork: That's correct. deprecated_version is an indication that the featurizer should no longer be used (which implies that it has already been released). Let me know if you can think of a less confusing name.

PR #139, opened 2020-02-11
Added ForwardFillImputerFeaturizer and BackwardFillImputerFeaturizer

davidbrownellWork:
These featurizers will be used by the TimeSeriesImputer refactor and are similar in nature to Mean-, Median-, MinMax-, and Mode-ImputerFeaturizers.


src/Featurizers/BackwardFillImputerFeaturizer.h
```diff
@@ -0,0 +1,164 @@
...
+        T const &                           value(TheseTraits::GetNullableValue(input));
+        std::uint32_t                       toWrite(_numPending + 1);
+
+        while(toWrite--)
+            callback(value);
+
+        _numPending = 0;
+    }
+
+    void flush_impl(typename BaseType::CallbackFunction const &) override {
```
angryjinyan: And where would this method be used? I didn't find an usage of this.

src/Tools/CodeGenerator/Featurizers.yaml
```diff
@@ -1,4 +1,59 @@
...
+        execute(NaN)
+        flush() // Exception, as pending values remain
+
+    has_dynamic_output: true
+    transformer_maintains_state: true
+
+    templates:
+      - name: T
+        types:
+          - int8
```
angryjinyan: I remember Dmitri suggested that it would be better to list these types sizewise, like int8, uint8, int16, uint16... Will there be another refactor for this?

src/Featurizers/BackwardFillImputerFeaturizer.h
yuslepukhin: This is called from a parent class

---
In reply to: [377925921](https://github.com/microsoft/FeaturizersLibrary/pull/139#discussion_r377925921) [](ancestors = 377925921)

src/Tools/CodeGenerator/Featurizers.yaml
yuslepukhin: The kernels have been refactored already.

---
In reply to: [377926952](https://github.com/microsoft/FeaturizersLibrary/pull/139#discussion_r377926952) [](ancestors = 377926952)

angryjinyan: Gotcha!

src/Featurizers/ForwardFillImputerFeaturizer.h
```diff
@@ -0,0 +1,88 @@
...
+
+    // MSVC runs into problems when the declaration and definition are separated
+    void execute_impl(typename BaseType::InputType const &input, typename BaseType::CallbackFunction const &callback) override {
+        // ----------------------------------------------------------------------
+        using TheseTraits                   = Traits<typename BaseType::InputType>;
+        // ----------------------------------------------------------------------
+
+        if(TheseTraits::IsNull(input)) {
+            if(_validLastValue == false)
+                throw std::runtime_error("No source value for forward fill");
```
yuslepukhin: >throw std::runtime_error("No source value for forward fill"); [](start = 16, length = 61)

Who is supposed to handle this or deal with this? Is this a caller problem or a data problem?

src/Featurizers/BackwardFillImputerFeaturizer.h
```diff
@@ -0,0 +1,164 @@
...
+    // |
+    // |  Public Methods
+    // |
+    // ----------------------------------------------------------------------
+    BackwardFillImputerTransformer(void);
+    BackwardFillImputerTransformer(Archive &ar);
+
+    FEATURIZER_MOVE_CONSTRUCTOR_ONLY(BackwardFillImputerTransformer);
+
+    bool has_pending(void) const;
```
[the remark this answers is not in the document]
yuslepukhin: Probably tests

---
In reply to: [377925299](https://github.com/microsoft/FeaturizersLibrary/pull/139#discussion_r377925299) [](ancestors = 377925299)

src/Featurizers/ForwardFillImputerFeaturizer.h
davidbrownellWork: This would be a data problem; it indicates that a null value has been received before a valid value, and there is nothing that can be used to populate this value. I considered avoiding the error case by making all output types optional (which would let us propagate null values for this particular case), but that seemed overly heavy. If we find that this is a common case, we can introduce more intelligent logic if the output happens to be a nullable type.

I'm going to add some comments in the code to describe this scenario.

---
In reply to: [377934541](https://github.com/microsoft/FeaturizersLibrary/pull/139#discussion_r377934541) [](ancestors = 377934541)

davidbrownellWork: Actually, now that I think about it, I'm going to add a constructor that let's a caller optionally provide a default value so that this scenario will be less common.

---
In reply to: [379533991](https://github.com/microsoft/FeaturizersLibrary/pull/139#discussion_r379533991) [](ancestors = 379533991,377934541)

PR #144, opened 2020-02-14
Broda/forward fill imputers

davidbrownellWork:


davidbrownellWork:
I apologize that I lost the comments associated with the previous PR - the PR can noisy after a particularly large rebase on master, so I closed it with the thought that I could reopen it once the rebase was complete. Unfortunately, it doesn't look like it works that way and I couldn't find an option to reopen the older one.

src/Tools/CodeGenerator/Featurizers.yaml
```diff
@@ -1,4 +1,59 @@
 featurizers:
+  # ----------------------------------------------------------------------
+  - name: BackwardFillImputerFeaturizer
+    estimator_name: BackwardFillImputerEstimator
```
michaelgsharp: This featurizer wont always return a row and needs flush, so `has_dynamic_output` is needed.

davidbrownellWork: > This featurizer wont always return a row and needs flush, so `has_dynamic_output` is needed.

That is specified below (line 31), but I will move it up so it is more visible.

src/Featurizers/BackwardFillImputerFeaturizer.h
```diff
@@ -0,0 +1,224 @@
...
+
+template <typename T>
+bool BackwardFillImputerTransformer<T>::operator==(BackwardFillImputerTransformer const &other) const {
+    if(_numPending != other._numPending)
+        return false;
+
+    // We aren't using nonstd::optional's comparison operator as we aren't able to disable compiler warnings
+    // associated with the comparison of float values as that code has already been #included by the time
+    // that we get here. Do things manually instead.
+    if(static_cast<bool>(_defaultValue) != static_cast<bool>(_defaultValue))
```
michaelgsharp: other._defaultValue

src/Featurizers/ForwardFillImputerFeaturizer.h
```diff
@@ -0,0 +1,185 @@
...
+    // MSVC runs into problems when the declaration and definition are separated
+    void execute_impl(typename BaseType::InputType const &input, typename BaseType::CallbackFunction const &callback) override {
+        // ----------------------------------------------------------------------
+        using TheseTraits                   = Traits<typename BaseType::InputType>;
+        // ----------------------------------------------------------------------
+
+        if(TheseTraits::IsNull(input)) {
+            if(_validLastValue == false) {
+                // This scenario will happen when the first imputed item is null, as we haven't seen a valid value yet.
+                throw std::runtime_error("No source value for forward fill");
```
michaelgsharp: Is there any way to suppress this error and let it continue? This is one of the use cases AutoML told us is the reasons that they will use multiple imputation strategies, so that if any of them still have nulls that another imputation strategy will catch it.

src/Featurizers/ForwardFillImputerFeaturizer.h
```diff
@@ -0,0 +1,185 @@
...
+void ForwardFillImputerTransformer<T>::save(Archive &ar) const /*override*/ {
+    // Version
+    Traits<std::uint16_t>::serialize(ar, 1); // Major
+    Traits<std::uint16_t>::serialize(ar, 0); // Minor
+
+    // Data
+    Traits<bool>::serialize(ar, _validLastValue);
+
+    if(_validLastValue)
+        Traits<T>::serialize(ar, _lastValue);
```
michaelgsharp: I think may have a potential bug here. I believe these are being serialized so that the default value will be preserved, is that correct? If so, if I run the transformer through some data first, the `_validLastValue` will be overwritten. Then, if I save the transformer after that data has been used, it will serialize a different value then what was originally provided. I think we should go with the approach that the backfill is using, with a different variable to store just the default value.

src/Featurizers/BackwardFillImputerFeaturizer.h
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    void flush_impl(typename BaseType::CallbackFunction const &callback) override {
+        // ----------------------------------------------------------------------
+        using NullableTypeTraits            = Traits<NullableType>;
+        // ----------------------------------------------------------------------
+
+        if(_numPending != 0) {
+            if(NullableTypeTraits::IsNull(_defaultValue))
+                throw std::runtime_error("Pending backward fill items remain");
```
michaelgsharp: Is there any way to suppress this error and let it continue? This is one of the use cases AutoML told us is the reasons that they will use multiple imputation strategies, so that if any of them still have nulls that another imputation strategy will catch it.

src/Featurizers/ForwardFillImputerFeaturizer.h
davidbrownellWork: There is, but the code path should only be exercised if the output type is optional and (possibly) they opt into the behavior. I'm going to add a TODO comment for now, but I want to spend some time thinking about the resulting complexity before implementing the functionality.

---
In reply to: [383480925](https://github.com/microsoft/FeaturizersLibrary/pull/144#discussion_r383480925) [](ancestors = 383480925)

src/Featurizers/BackwardFillImputerFeaturizer.h
davidbrownellWork: Similar comment as the other - I've added a TODO but we need to think about the scenario more before implementing the solution.

---
In reply to: [383484002](https://github.com/microsoft/FeaturizersLibrary/pull/144#discussion_r383484002) [](ancestors = 383484002)

src/Featurizers/ForwardFillImputerFeaturizer.h
davidbrownellWork: Per our conversation, I've updated serialization and equality to only operator on initial state (and ignore working state).

PR #150, opened 2020-02-21
ShortGrainDropperFeaturizer

wangyems:
ShortGrainDropper hack version:
1, vector<string> currently not supported in codegen
2, consider optimizing by using vector of strings in unordered_set rather than string due to string copy in Traits<>::ToString() 
3, need more info on validation for minPoints generation


src/Featurizers/ShortGrainDropperFeaturizer.h
```diff
@@ -0,0 +1,255 @@
...
+// |
+// |  Implementation
+// |
+// ----------------------------------------------------------------------
+// ----------------------------------------------------------------------
+// ----------------------------------------------------------------------
+
+// ----------------------------------------------------------------------
+// |
+// |  ShortGrainDropperTransformer
```
davidbrownellWork: >ShortGrainDropperTransformer [](start = 6, length = 28)

Move the implementation of this class into a .cpp file as it isn't a template and therefore doesn't need to be in the header file.

src/Featurizers/ShortGrainDropperFeaturizer.h
```diff
@@ -0,0 +1,255 @@
...
+    GrainsSet const                         _grainsToDrop;
+
+    // ----------------------------------------------------------------------
+    // |
+    // |  Private Methods
+    // |
+    // ----------------------------------------------------------------------
+     // MSVC has problems when the declaration and definition are separated
+    void execute_impl(typename BaseType::InputType const &input, typename BaseType::CallbackFunction const &callback) override {
+        callback(_grainsToDrop.find(Traits<decltype(input)>::ToString(input)) != _grainsToDrop.end());
```
davidbrownellWork: >Traits<decltype(input)>::ToString(input) [](start = 36, length = 40)

I don't think that we need to convert the input to a string here, as the vector of strings should be used to lookup membership in _grainsToDrop.

src/Featurizers/ShortGrainDropperFeaturizer.h
```diff
@@ -0,0 +1,255 @@
...
+    // |  Public Methods
+    // |
+    // ----------------------------------------------------------------------
+    ShortGrainDropperEstimator(
+        AnnotationMapsPtr pAllColumnAnnotations,
+        size_t colIndex,
+        std::uint32_t windowSize,
+        std::vector<std::uint32_t> lags,
+        std::uint32_t maxHorizon,
+        nonstd::optional<std::uint32_t> cv
```
davidbrownellWork: I don't know what these variable names mean. Consider making the names more descriptive or add comments so that the caller has an idea of what to expect.

src/Featurizers/ShortGrainDropperFeaturizer.h
```diff
@@ -0,0 +1,255 @@
...
+                ++grainsMapIter->second;
+            else
+                groupByGrains.emplace(pBufferStr, 1);
+            ++pBuffer;
+        }
+
+        for (GrainsMap::value_type const & groupByGrainsElement : groupByGrains) {
+            if (groupByGrainsElement.second <= _minPoints)
+                _grainsToDrop.emplace(std::move(groupByGrainsElement.first));
+        }
```
davidbrownellWork: This logic doesn't feel right to me, as it seems like we need to maintain the count across all training and only calculate the grains to drop once training is complete.

src/Featurizers/ShortGrainDropperFeaturizer.h
```diff
@@ -0,0 +1,255 @@
...
+///
+class ShortGrainDropperTransformer : public StandardTransformer<std::vector<std::string>, bool> {
+public:
+    // ----------------------------------------------------------------------
+    // |
+    // |  Public Types
+    // |
+    // ----------------------------------------------------------------------
+    using BaseType                          = StandardTransformer<std::vector<std::string>, bool>;
+    using GrainsSet                         = std::unordered_set<std::string>;
```
davidbrownellWork: ><std::string [](start = 64, length = 12)

Should this be `std::vector<std::string>` given that a single grain may be comprised of multiple strings?

src/Featurizers/ShortGrainDropperFeaturizer.h
```diff
@@ -0,0 +1,255 @@
...
+    // ----------------------------------------------------------------------
+     // MSVC has problems when the declaration and definition are separated
+    void execute_impl(typename BaseType::InputType const &input, typename BaseType::CallbackFunction const &callback) override {
+        callback(_grainsToDrop.find(Traits<decltype(input)>::ToString(input)) != _grainsToDrop.end());
+    }
+};
+
+/////////////////////////////////////////////////////////////////////////
+///  \class         ShortGrainDropperEstimator
+///  \brief         Estimator to determine which grain to drop
```
davidbrownellWork: >Estimator to determine which grain to drop [](start = 20, length = 42)

This comment can be improved. Based on the ctor parameters, I gather that the logic used to determine if a grain should be dropped is based on time. However, I don't see anything in the input related to time.

src/Featurizers/ShortGrainDropperFeaturizer.h
wangyems: The logic here is exactly calculating the grains to drop when training is complete, as the signature of fit_impl() is fit_impl(typename BaseType::InputType const *pBuffer, size_t cElements)

src/Featurizers/ShortGrainDropperFeaturizer.h
wangyems: it looks like the way to determine which grain to drop is not directly based on time

src/Featurizers/ShortGrainDropperFeaturizer.h
```diff
@@ -0,0 +1,230 @@
...
+/////////////////////////////////////////////////////////////////////////
+///  \class         ContainerHash
+///  \brief         Hash function for Container type
+///
+template <typename Container>
+struct ContainerHash {
+    std::size_t operator()(Container const& container) const noexcept {
+        std::size_t hash = 0;
+        for (typename Container::value_type const & val : container)
+            hash ^= std::hash<typename Container::value_type>{}(val) + 0x9e3779b9 + (hash << 6) + (hash >> 2);
```
davidbrownellWork: >hash ^= std::hash<typename Container::value_type>{}(val) + 0x9e3779b9 + (hash << 6) + (hash >> 2); [](start = 12, length = 98)

This represents too much hard coded logic. Either:

1) Continue to use an std::unordered_set and use MurmurHash
2) Use std::set

src/Featurizers/ShortGrainDropperFeaturizer.h
```diff
@@ -0,0 +1,230 @@
...
+    // |  Private Methods
+    // |
+    // ----------------------------------------------------------------------
+    void execute_impl(typename BaseType::InputType const &input, typename BaseType::CallbackFunction const &callback) override;
+};
+
+/////////////////////////////////////////////////////////////////////////
+///  \class         ShortGrainDropperEstimator
+///  \brief         Estimator to determine which grain to drop given the
+///                 threshod minPoints calculated by:
```
davidbrownellWork: I'm trying to read this from the perspective of a new developer as I write the following comments....

src/Featurizers/ShortGrainDropperFeaturizer.h
```diff
@@ -0,0 +1,230 @@
...
+///
+///                 windowSize: the rolling window size.
+///                 lags: The lag sizes.
```
davidbrownellWork: >The lag sizes. [](start = 26, length = 14)

What is a lag?

src/Featurizers/ShortGrainDropperFeaturizer.h
```diff
@@ -0,0 +1,230 @@
...
+    // ----------------------------------------------------------------------
+    void execute_impl(typename BaseType::InputType const &input, typename BaseType::CallbackFunction const &callback) override;
+};
+
+/////////////////////////////////////////////////////////////////////////
+///  \class         ShortGrainDropperEstimator
+///  \brief         Estimator to determine which grain to drop given the
+///                 threshod minPoints calculated by:
+///
+///                 windowSize: the rolling window size.
```
davidbrownellWork: >the rolling window size. [](start = 32, length = 24)

Based on the description, I'm not sure how a rolling window is used when dropping grains. There seems to be a dependency on prior featurizers that isn't clear based on the class name. For example, it doesn't seem like we could use this featurizers to drop any content that happens to have a grain.

src/Featurizers/ShortGrainDropperFeaturizer.h
```diff
@@ -0,0 +1,230 @@
...
+///                 lags: The lag sizes.
+///                 maxHorizon: the desired length of forecasting (the number of new rows generated)
```
davidbrownellWork: >the desired length of forecasting (the number of new rows generated) [](start = 32, length = 68)

Is this the minimum number of entries that must be encountered to prevent a grain from being dropped? The term "horizon" doesn't fit my mental model of what this class is doing.

src/Featurizers/ShortGrainDropperFeaturizer.h
```diff
@@ -0,0 +1,230 @@
...
+///                 cv: the number of cross validations.
```
davidbrownellWork: >the number of cross validations. [](start = 24, length = 32)

What are cross validations?

src/Featurizers/ShortGrainDropperFeaturizer.h
davidbrownellWork: >windowSize: the rolling window size. [](start = 20, length = 36)

I would expect these parameters to be documented along with the constructor rather than at the class level.

src/Featurizers/ShortGrainDropperFeaturizer.h
davidbrownellWork: Let's discuss this in person to make sure that we are on the same page.

---
In reply to: [384062585](https://github.com/microsoft/FeaturizersLibrary/pull/150#discussion_r384062585) [](ancestors = 384062585)

src/Traits.h
```diff
@@ -131,6 +133,37 @@ inline bool IsValid(TypeId id) {
...
+    MurmurHash3_x86_32(reinterpret_cast<unsigned char const*>(&value), sizeof(value), seed, &hash);
+    return hash;
+}
+
+/////////////////////////////////////////////////////////////////////////
+///  \class         ContainerHash
+///  \brief         Hash function for Container type
+///
+template <typename Container>
+struct ContainerHash {
```
davidbrownellWork: >struct ContainerHash [](start = 0, length = 20)

There should be tests in Traits_UnitTest.cpp that exercise this functionality.

src/Featurizers/ShortGrainDropperFeaturizer.h
```diff
@@ -0,0 +1,216 @@
...
+// ----------------------------------------------------------------------
+// ----------------------------------------------------------------------
+template <size_t MaxNumTrainingItemsV>
+bool ShortGrainDropperEstimator<MaxNumTrainingItemsV>::begin_training_impl(void) /*override*/ {
+    return true;
+}
+
+template <size_t MaxNumTrainingItemsV>
+void ShortGrainDropperEstimator<MaxNumTrainingItemsV>::complete_training_impl(void) /*override*/ {
+    for (GrainsMap::value_type const & groupByGrainsElement : _groupByGrains) {
```
davidbrownellWork: >_groupByGrains [](start = 62, length = 14)

We don't need to keep _groupByGrains once _grainsToDrop is complete; consider clearing _groupByGrains at the end of this loop.

src/Featurizers/UnitTests/ShortGrainDropperFeaturizer_UnitTests.cpp
```diff
@@ -0,0 +1,190 @@
...
+        inferencingInput,
+        inferencingOutput,
+        windowSize,
+        lags,
+        maxHorizon,
+        cv
+    );
+}
+
+TEST_CASE("Standard Test 2") {
```
davidbrownellWork: >2 [](start = 25, length = 1)

How is this test different from "Standard Test"? Consider a more descriptive name for the test.

src/Featurizers/UnitTests/ShortGrainDropperFeaturizer_UnitTests.cpp
```diff
@@ -0,0 +1,190 @@
...
+#define CATCH_CONFIG_MAIN
+#include "catch.hpp"
+
+#include "../../Featurizers/ShortGrainDropperFeaturizer.h"
+#include "../TestHelpers.h"
+#include "../../Archive.h"
+
+namespace NS = Microsoft::Featurizer;
+
+void TestImpl(std::vector<std::vector<std::vector<std::string>>> trainingBatches,
```
davidbrownellWork: >TestImpl [](start = 5, length = 8)

I'd expect to see more tests that exercise the estimator's parameters. For example, consider writing a series of tests that illustrate how results change with the same training set data when different parameters are used. The parameters should be selected to exercise a variety of different scenarios. Test SECTIONs will make the implementation a bit easier ([https://github.com/catchorg/Catch2/blob/master/docs/tutorial.md#test-cases-and-sections](https://github.com/catchorg/Catch2/blob/master/docs/tutorial.md#test-cases-and-sections))

PR #151, opened 2020-02-21
Initial implementation of min, max, mean, and sum calculators

michaelgsharp:
This is the initial implementation of the min, max, mean, and sum calculators and their associated unit tests. 

Still need to decide exactly how we want to handle NaN values. Currently, if a NaN value is passed in, the output becomes NaN. I think ignoring NaN would be better, but wanted to discuss it so for now just creating this PR as is.


src/Featurizers/Calculators/SumCalculator.h
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+private:
+    friend class CalculatorBase<SumCalculator<T>, T, T>;
+
+    T const                                 _initialValue;
+    T                                       _sum;
+
+    inline void update_impl(T value) {
+        _sum += value;
```
angryjinyan: value should be validated before addition

src/Featurizers/Calculators/MinCalculator.h
```diff
@@ -0,0 +1,54 @@
...
+        this->reset();
+    }
+
+private:
+    friend class CalculatorBase<MinCalculator<T>, T, T>;
+
+    T                                       _minValue;
+
+    inline void update_impl(T value) {
+        if(value < _minValue) {
```
angryjinyan: value should be validated

src/Featurizers/Calculators/MaxCalculator.h
```diff
@@ -0,0 +1,54 @@
...
+        this->reset();
+    }
+
+private:
+    friend class CalculatorBase<MaxCalculator<T>, T, T>;
+
+    T                                       _maxValue;
+
+    inline void update_impl(T value) {
+        if(value > _maxValue) {
```
angryjinyan: value should be validated

src/Featurizers/Calculators/MaxCalculator.h
```diff
@@ -0,0 +1,54 @@
...
+            _maxValue = value;
+        }
+    }
+
+    inline T commit_impl(void) {
+        return _maxValue;
```
angryjinyan: I think a flag should be added to mark if there are valid input values passed in. Refer to _has_valid_input in NormUpdaters.h. After adding validation for update_impl, commit_impl can return std::numeric_limits<T>::min() without valid input values.

src/Featurizers/Calculators/UnitTests/MaxCalculator_UnitTest.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+    calc.reset();
+    CHECK(calc.commit() == std::numeric_limits<int>::min());
+
+    calc.update(10);
+    CHECK(calc.commit() == 10);
+}
+
+TEST_CASE("Standard - double") {
+    NS::MaxCalculator<double>                   calc;
```
angryjinyan: maybe add unit tests for NaNs?

PR #154, opened 2020-02-27
Circular Buffer and Iterator 

michaelgsharp:
This PR adds in a circular buffer, a custom container class created for operations with shifted windows that avoids allocating new memory, and a new forward iterator to go with the buffer.

The circular buffer uses a vector as its underlying storage, reserves the memory once, and then never allocates more. When the vector is full, the buffer "circles" back to the beginning and overwrites the oldest value. It provides `begin()` and `end()` iterator methods, as well as a `range()` method to get N values starting from a given offset.

The iterator handles and "circling" that needs to happen so the end user doesn't have to know or care about it. If the iterator reaches the last position of the vector, it will automatically wrap back to the first one. It still provides the ability to know when its at the `end` of the data by knowing how many times it is allowed to be incremented. This means the normal `while(start_iter != end_iter)` syntax still works as expected.


src/Featurizers/Components/WindowFeaturizerBase.h
```diff
@@ -0,0 +1,244 @@
...
+        // Current number of element held by this container. It start from 0 and grows to _max_size
+        size_t                                  _cur_size;
+
+        // The vector that holds the actual data. This is set to hold _max_size elements and then
+        // is never resized.
+        std::vector<T>                          _data;
+
+    public:
+        // ----------------------------------------------------------------------
+        // |  Public Data
```
davidbrownellWork: >Data [](start = 21, length = 4)

Types

src/Featurizers/Components/WindowFeaturizerBase.h
```diff
@@ -0,0 +1,244 @@
...
+        typedef CircularIterator<T>             iterator;
+        typedef CircularIterator<const T>       const_iterator;
```
davidbrownellWork: Prefer using statements over typedefs.

src/Featurizers/Components/WindowFeaturizerBase.h
```diff
@@ -0,0 +1,244 @@
...
+
+        // ----------------------------------------------------------------------
+        // |  Public Methods
+        CircularBuffer(size_t max_size) : _max_size(max_size), _start_offset(0),  _cur_size(0) {
```
davidbrownellWork: >max_size [](start = 30, length = 8)

Raise error if max_size == 0

davidbrownellWork: >_cur_size [](start = 82, length = 9)

This variable isn't necessary, as _data.size() can be used directly.

src/Featurizers/Components/WindowFeaturizerBase.h
```diff
@@ -0,0 +1,244 @@
...
+            _data.reserve(_max_size);
+        }
+
+        iterator begin() {
+            // When _cur_size is 0 we want to return an "end" iterator.
+            if (_cur_size == 0) {
+                return iterator();
+            }
+            return iterator(&_data[0], _max_size, _cur_size, _start_offset);
```
davidbrownellWork: >&_data[0] [](start = 28, length = 9)

Nit: _data.data()

src/Featurizers/Components/WindowFeaturizerBase.h
```diff
@@ -0,0 +1,244 @@
...
+        }
+
+        const_iterator begin() const {
+            return begin();
```
davidbrownellWork: >return begin(); [](start = 12, length = 15)

I'm not sure how this code is working. The type CircularIterator<T> should not be convertible to CircularIterator<T const>.

src/Featurizers/Components/WindowFeaturizerBase.h
```diff
@@ -0,0 +1,244 @@
...
+                _data.emplace_back(std::move(value));
+                ++_cur_size;
+            }
+        }
+
+        // provide a pair of begin and end iterator for the n elements requested
+        std::tuple<iterator, iterator> range(size_t n, size_t offset=0) {
+            // Since this class is used for window operations only
+            // number of requested elements is never bigger than the max_size
+            assert(n + offset <= _max_size);
```
davidbrownellWork: >assert(n + offset <= _max_size); [](start = 12, length = 32)

This assert will only be active for debug builds. This should be an exception instead and there should be a test that exercises the functionality.

src/Featurizers/Components/WindowFeaturizerBase.h
michaelgsharp: We did the typedef cause in your email to us you said we needed to provide typedefs. Would `using` be better? Or did we misunderstand and put the typedefs in the wrong spot?

src/Featurizers/Components/WindowFeaturizerBase.h
michaelgsharp: Good catch. Fixed.

src/Featurizers/Components/WindowFeaturizerBase.h
michaelgsharp: Added in `std::invalid_argument` exception there.

src/Featurizers/Components/WindowFeaturizerBase.h
angryjinyan: Do you mean convert to const CircularIterator? There's a method in the CircularIterator class that can do the conversion.

src/Featurizers/Components/WindowFeaturizerBase.h
davidbrownellWork: My bad. I am using old terminology with new concepts. I should have said "we should have alias types for iterator and const_iterator". There are 2 ways of creating alias types in C++ - 1) via typedef (old with corner cases) 2) via using (new without corner cases).

---
In reply to: [385338421](https://github.com/microsoft/FeaturizersLibrary/pull/154#discussion_r385338421) [](ancestors = 385338421)

src/Featurizers/Components/WindowFeaturizerBase.h
```diff
@@ -0,0 +1,244 @@
...
+        // |  Public Methods
+
+        // Default constructor, already at the "end".
+        CircularIterator()
+            : _itr(nullptr), _size(0), _cur_index(0),
+            _max_increments(0), _cur_increment(0) {
+        }
+
+        CircularIterator(T* T, size_t container_max_size, size_t max_increments, size_t starting_offset = 0)
+            : _itr(T), _size(container_max_size), _cur_index(starting_offset),
```
davidbrownellWork: >_itr [](start = 14, length = 4)

Raise exceptions if t is nullptr, container_max_size == 0, max_increments == 0, or starting_offset > container_max_size. Also raise error if max_increments > container_max_size. Add corresponding tests.

src/Featurizers/Components/WindowFeaturizerBase.h
```diff
@@ -0,0 +1,244 @@
...
+            _max_increments(max_increments), _cur_increment(0) {
+        }
+
+        // Pre-increment operator.
+        CircularIterator& operator++ () {
+            assert(++_cur_increment <= _max_increments);
```
davidbrownellWork: >assert(++_cur_increment <= _max_increments); [](start = 12, length = 44)

raise an exception

src/Featurizers/Components/WindowFeaturizerBase.h
michaelgsharp: removed `_cur_size` and replaced its use with `_data.size()`

src/Featurizers/Components/WindowFeaturizerBase.h
```diff
@@ -0,0 +1,244 @@
...
+            ++_cur_index;
+            _cur_index %= _size;
+            return *this;
+        }
+
+        // Post-increment operator.
+        CircularIterator operator++ (int) {
+            CircularIterator tmp(*this);
+
+            assert(++_cur_increment <= _max_increments);
```
davidbrownellWork: >assert(++_cur_increment <= _max_increments); [](start = 12, length = 44)

Raise an exception.

src/Featurizers/Components/WindowFeaturizerBase.h
michaelgsharp: Changed `&_data[0]` to `_data.data()`

src/Featurizers/Components/WindowFeaturizerBase.h
```diff
@@ -0,0 +1,244 @@
...
+
+            ++_cur_index;
+            _cur_index %= _size;
```
davidbrownellWork: Rather than duplicating this code, you can call operator++ here.

src/Featurizers/Components/WindowFeaturizerBase.h
```diff
@@ -0,0 +1,244 @@
...
+            return tmp;
+        }
+
+        // Two-way comparison: v.begin() == v.cbegin() and vice versa.
+        template<class OtherType>
```
davidbrownellWork: > template<class OtherType> [](start = 7, length = 26)

By making this a template, we are allowing comparison between different iterator types. I'd recommend removing the template parameter to avoid potential wonkiness.

src/Featurizers/Components/WindowFeaturizerBase.h
michaelgsharp: Good to know. We used an assert because we tested with a normal std::iterator and it uses an assert, but if the assert is only active for debug then its not as helpful. We did verify it works, but since we can't "catch" an assert in the test we had to remove the test for the build. Ill make the change and add the test back in.

src/Featurizers/Components/WindowFeaturizerBase.h
michaelgsharp: we use max_increments = 0 with our "end" iterator, so we can't raise an exception on that, but I will add in the others.

src/Featurizers/Components/WindowFeaturizerBase.h
```diff
@@ -0,0 +1,244 @@
...
+        bool operator == (const CircularIterator<OtherType>& rhs) const {
+            return _itr == rhs._itr && _cur_index == rhs._cur_index &&
+                !((_cur_increment == _max_increments) ^ (rhs._cur_increment == rhs._max_increments));
```
davidbrownellWork: >!((_cur_increment == _max_increments) ^ (rhs._cur_increment == rhs._max_increments)) [](start = 16, length = 84)

I've spent 2 minutes looking at this code figuring out what it is doing and corner cases. Could it be simplified by comparing the 2 object's _cur_increment and _max_increment values like what is being done for _iter and _cur_index?

src/Featurizers/Components/WindowFeaturizerBase.h
```diff
@@ -0,0 +1,244 @@
...
+            assert(++_cur_increment <= _max_increments);
+
+            ++_cur_index;
+            _cur_index %= _size;
+            return tmp;
+        }
+
+        // Two-way comparison: v.begin() == v.cbegin() and vice versa.
+        template<class OtherType>
+        bool operator == (const CircularIterator<OtherType>& rhs) const {
```
davidbrownellWork: >bool operator == (const CircularIterator<OtherType>& rhs) const { [](start = 8, length = 65)

There is a subtle bug here in that a very old iterator could compare as equal to a new iterator after the internal memory buffer has wrapped. I would recommend adding a version value that is set in the iterator's ctor and checking that as well. In the CircularBuffer class, update that version value every time an element is added.

src/Featurizers/Components/WindowFeaturizerBase.h
angryjinyan: This code is to check whether both iterators are end iterators and two end iterators can have different _cur_increment and _max_increment

src/Featurizers/Components/WindowFeaturizerBase.h
michaelgsharp: Also, is there a problem if max_increments is greater than max_size? That would allow them to loop over the data multiple times if desired. Or do you think thats bad behavior so we just shoudln't allow it?

src/Featurizers/Components/WindowFeaturizerBase.h
```diff
@@ -0,0 +1,244 @@
...
+
+        T* operator-> () const {
+            assert(_cur_increment != _max_increments);
+            return (_itr+_cur_index);
+        }
+
+        // One way conversion: iterator -> const_iterator
+        operator CircularIterator<const T>() const {
+            return CircularIterator<const T>(_itr);
+        }
```
davidbrownellWork: I'd recommend removing this. If you really want to be able to convert to a const iterator:

1) Create a method named ToConstIterator so callers can opt in to the functionality (we don't want the compiler to do this for us)
2) Only enable the functionality when T is not const.

src/Featurizers/Components/WindowFeaturizerBase.h
```diff
@@ -0,0 +1,244 @@
...
+                !((_cur_increment == _max_increments) ^ (rhs._cur_increment == rhs._max_increments));
+        }
+
+        template<class OtherType>
+        bool operator != (const CircularIterator<OtherType>& rhs) const {
+            return !operator==(rhs);
+        }
+
+        T& operator* () const {
+            assert(_cur_increment != _max_increments);
```
davidbrownellWork: >assert(_cur_increment != _max_increments); [](start = 12, length = 42)

I don't think that we want to throw exceptions rather than asserts everywhere in the code. Let's chat if you feel like that would add too much runtime overhead.

src/Featurizers/Components/WindowFeaturizerBase.h
michaelgsharp: Exception added

src/Featurizers/Components/WindowFeaturizerBase.h
michaelgsharp: Exception added.

src/Featurizers/Components/WindowFeaturizerBase.h
michaelgsharp: Refactored to call ++ instead.

src/Featurizers/Components/WindowFeaturizerBase.h
michaelgsharp: Good to know. I didn't catch on to that. I have removed the template.

src/Featurizers/Components/WindowFeaturizerBase.h
davidbrownellWork: I'd recommend adding comments if this is a scenario that you are supporting.

---
In reply to: [385345847](https://github.com/microsoft/FeaturizersLibrary/pull/154#discussion_r385345847) [](ancestors = 385345847)

src/Featurizers/Components/WindowFeaturizerBase.h
michaelgsharp: Can you explain more on this? I am not sure I follow the situation.

src/Featurizers/Components/WindowFeaturizerBase.h
michaelgsharp: Per our offline discussion. Leaving as asserts for now.

src/Featurizers/Components/WindowFeaturizerBase.h
michaelgsharp: Replaced the `typedefs` with `using`

src/Featurizers/Components/WindowFeaturizerBase.h
michaelgsharp: Resolving per offline discussion.

src/Featurizers/Components/WindowFeaturizerBase.h
michaelgsharp: Resolving per offline discussions. Following STL templates and old iterators are invalid when the underlying data is changed.

src/Featurizers/Components/WindowFeaturizerBase.h
michaelgsharp: Added an exception to make sure `max_increments` is not greater than `max_size`

src/Featurizers/Components/WindowFeaturizerBase.h
michaelgsharp: Made the changes discussed offline. Removed the conversion code and added `_impl` methods.

src/Featurizers/Components/WindowFeaturizerBase.h
michaelgsharp: Made the changes discussed offline. Removed the conversion code and added _impl methods.

src/Featurizers/Components/WindowFeaturizerBase.h
michaelgsharp: Simplified the equality logic by removing the `nor`.

src/Featurizers/Components/WindowFeaturizerBase.h
```diff
@@ -0,0 +1,305 @@
...
+        // The number of elements that the underlying data structure has.
+        size_t                                  _size;
+
+        // The index this iterator is pointer it in the underlying data structure.
+        size_t                                  _cur_index;
+
+        // The maximum number of times this iterator can be incremented. Since this iterator will
+        // wrap at the end of the data structure, there is no real "end" of the data so this is
+        // used to determine when we have met the "end".
+        size_t                                  _max_increments;
```
davidbrownellWork: >size_t [](start = 8, length = 6)

This can be const.

src/Featurizers/Components/WindowFeaturizerBase.h
```diff
@@ -0,0 +1,305 @@
...
+        // |
+        // |  Private Data
+        // |
+        // ----------------------------------------------------------------------
+
+        // Pointer to the first element of the underlying data structure.
+        T*                                      _itr;
+
+        // The number of elements that the underlying data structure has.
+        size_t                                  _size;
```
davidbrownellWork: >size_t [](start = 8, length = 6)

This can be const.

src/Featurizers/Components/WindowFeaturizerBase.h
```diff
@@ -0,0 +1,305 @@
...
+    ///                 would move it past the end of the underlying data structure,
+    ///                 it will instead wrap around back to the beginning. This follows
+    ///                 STL iterators in that if the underlying data structure is changed,
+    ///                 the iterator is invalid and behavior is undefined if used.
+    ///                 This class is designed for use with the WindowFeaturizerBase and
+    ///                 so it uses assert and not exception. If this class needs to be
+    ///                 used elsewhere at a later date, then this will need to be addressed.
+    ///
+    template <class T>
+    class CircularIterator
```
davidbrownellWork: >CircularIterator [](start = 10, length = 16)

Can you separate the declarations and definitions for this class.

src/Featurizers/Components/WindowFeaturizerBase.h
michaelgsharp: Added const.

src/Featurizers/Components/WindowFeaturizerBase.h
michaelgsharp: Added const.

PR #157, opened 2020-03-04
Yaml additions for the RollingWindowFeaturizer

michaelgsharp:
Add in the Yaml changes for the RollingWindowFeaturizer. Since all calculations can return a NaN value, all calculations return a double value so only one featurizer is needed.


michaelgsharp:
I named them as `SimpleRollingWindowFeaturizer` and then for the one that only returns doubles I named it `AnalyticalRollingWindowFeaturizer`.

src/Tools/CodeGenerator/Featurizers.yaml
```diff
@@ -1,4 +1,101 @@
...
+            - int64
+            - uint8
+            - uint16
+            - uint32
+            - uint64
+            - float
+            - double
+
+    type_mappings:
+      - input_type: T
```
davidbrownellWork: >T [](start = 20, length = 1)

Should this be a tuple as described on line 11?

michaelgsharp: Line 11 I was trying to explain the input to the execute method. This type T is the template type. Would there be a better way for me to make that more clear? Or is this why you were talking about the execute method parameters with me?

davidbrownellWork: I think that the input type will end up being `tuple<datetime, vector<string>, horizon, T>`. The output will be `T`, where the output type is flagged as optional.

michaelgsharp: Is horizon going to be a type as well? I have it listed as uint8 now but can update and push again if needed.

davidbrownellWork: The horizon value will not be a template, but we will need to to be the same across MaxHorizon, RollingWindow, and LagLead. Let's make it uint16.

michaelgsharp:
@davidbrownellWork @angryjinyan I have made the updates as I understood them based on our meetings and discussions. If you could take another look I would appreciate it.

Also, we need to decide the order of the items in the vector. Is it going to be largest to smallest (3,2,1) or smallest to largest (1,2,3). If we aren't consistent then we will have problems.

angryjinyan:
I think the order needs to be 3, 2, 1 as this is the chronological order that 3 is the largest offset we look back.

michaelgsharp:
I agree with your order, I just wanted to make sure we were all on the same page because we could implement it either way and it woudln't really affect our algorithms.

src/Tools/CodeGenerator/Featurizers.yaml
```diff
@@ -1,4 +1,95 @@
...
+            |time(YYYY-MM-DD)| grain     | target| target_max        |
+            +================+===========+=======+===================+
+            | 2017-01-01     | A         | 10    | [NAN, NAN]        |
+            +----------------+-----------+-------+-------------------+
+            | 2017-02-01     | A         | 4     | [NAN, NAN]        |
+            +----------------+-----------+-------+-------------------+
+            | 2017-03-01     | A         | 6     | [NAN, 7]          |
+            +----------------+-----------+-------+-------------------+
+            | 2017-04-01     | A         | 11    | [7, 5]            |
+            +----------------+-----------+-------+-------------------+
```
angryjinyan: I think another example can be added here with window size is not the same as min window periods.

src/Tools/CodeGenerator/Featurizers.yaml
```diff
@@ -1,4 +1,95 @@
...
+        is_optional: false
+        description: The minimum number of elements that must be in an window to perform the calculation. If there is less, NaN will be returned.
+        
+      - type: AnalyticalRollingWindowCalculation
+        name: windowCalculation
+        is_optional: false
+        description: What calculation is going to be performed over the window data.
+
+      - type: uint32
+        name: maxHorizon
```
angryjinyan: Since we don't have horizon value in the execution() input data, do we still need the "max" prefix? There will be just one horizon all over the place.

src/Tools/CodeGenerator/Featurizers.yaml
michaelgsharp: The first example is that way, window size is 2 but it starts doing the math as soon as 1 row is available. Though looking again at my yaml I have the minWindowCount set to required. It should be optional and defaults to just 1. I will update it with that change.

src/Tools/CodeGenerator/Featurizers.yaml
michaelgsharp: I can remove the max prefix. I am fine either way. Mentally max makes sense to me, but I think I am the outsider with that. I will remove it.

src/Tools/CodeGenerator/Featurizers.yaml
```diff
@@ -1,4 +1,95 @@
...
+    release_version: 0.4.0
+    num_output_columns: 1
+    has_dynamic_output: false
+    description: |-
+        Calculates data based on a rolling window. Currently supports mean. Works for general time series data sets.
+        
+        Input type for this featurizer is a tuple of time point of current row, the grain columns, and target value column.
+
+        C++-style pseudo signature:
+        template <typename T> std::vector<double> execute(std::tuple<std::chrono::system_clock::time_point, std::vector<std::string>, T> const &value);
```
angryjinyan: add const & for tuple element

michaelgsharp: Added in const ref

src/Tools/CodeGenerator/Featurizers.yaml
michaelgsharp: Per the discussion we had, making this uint32.

michaelgsharp:
I have fixed the current comments @davidbrownellWork and @angryjinyan. If you could take another look when you have time I would appreciate it.

src/Tools/CodeGenerator/Featurizers.yaml
```diff
@@ -1,4 +1,95 @@
...
+  # ----------------------------------------------------------------------
+  - name: AnalyticalRollingWindowFeaturizer
+    estimator_name: AnalyticalRollingWindowEstimator
+    release_version: 0.4.0
+    num_output_columns: 1
+    has_dynamic_output: false
+    description: |-
+        Calculates data based on a rolling window. Currently supports mean. Works for general time series data sets.
+        
+        Input type for this featurizer is a tuple of time point of current row, the grain columns, and target value column.
```
davidbrownellWork: >Input type for this featurizer is a tuple of time point of current row, the grain columns, and target value column. [](start = 8, length = 115)

This description needs to be updated now that we are no longer taking a datetime as input.

src/Tools/CodeGenerator/Featurizers.yaml
```diff
@@ -1,4 +1,95 @@
...
+            |time(YYYY-MM-DD)| grain     | target| target_mean       |
+            +================+===========+=======+===================+
+            | 2017-01-01     | A         | 10    | [NAN]             |
+            +----------------+-----------+-------+-------------------+
+            | 2017-02-01     | A         | 4     | [10]              |
+            +----------------+-----------+-------+-------------------+
+            | 2017-03-01     | A         | 6     | [7]               |
+            +----------------+-----------+-------+-------------------+
+            | 2017-04-01     | A         | 11    | [5]               |
+            +----------------+-----------+-------+-------------------+
```
davidbrownellWork: Updates these tables now that we are no longer taking a datetime as input.

src/Tools/CodeGenerator/Featurizers.yaml
```diff
@@ -1137,6 +1228,99 @@ featurizers:
...
+  # ----------------------------------------------------------------------
+  - name: SimpleRollingWindowFeaturizer
+    estimator_name: SimpleRollingWindowEstimator
+    release_version: 0.4.0
+    num_output_columns: 1
+    has_dynamic_output: false
+    description: |-
+      Calculates data based on a rolling window. Currently supports minimum and maximum. Works for general time series data sets.
+      
+      Input type for this featurizer is a tuple of time point of current row, the grain columns, and target column to find the value.
```
davidbrownellWork: >time point of current row [](start = 51, length = 25)

Update this desc and tables.

src/Tools/CodeGenerator/Featurizers.yaml
misharp: Good catch. I have removed that from the description and added that we are assuming the data is already sorted in the correct order.

michaelgsharp: Good catch. I have removed that from the description and added that we are assuing the data is already sorted in the correct order.

src/Tools/CodeGenerator/Featurizers.yaml
michaelgsharp: Removed the date column.

src/Tools/CodeGenerator/Featurizers.yaml
michaelgsharp: Updated

PR #159, opened 2020-03-05
Yaml addition for MaxHorizonFeaturizer

gh-yewang:


src/Tools/CodeGenerator/Featurizers.yaml
```diff
@@ -639,6 +639,72 @@ featurizers:
       - input_type: TInputDouble
         output_type: double
 
+    # ----------------------------------------------------------------------
+  - name: MaxHorizonFeaturizer
+    estimator_name: MaxHorizonFeaturizer
```
davidbrownellWork: Make sure that you run `Tools/CodeGenerator/Build.py` on this before created a PR. The changes aren't valid as they are currently implemented.

davidbrownellWork (changes requested):

src/Tools/CodeGenerator/Featurizers.yaml
```diff
@@ -639,6 +639,72 @@ featurizers:
...
+          +------------+-------+-------+----------------+
+          | 2017-03-01 | A     | 2     | 2              |
+          +------------+-------+-------+----------------+
+          | 2017-04-01 | A     | 3     | 1              |
+          +------------+-------+-------+----------------+
+          | 2017-04-01 | A     | 3     | 2              |
+          +------------+-------+-------+----------------+
+
+    configuration_params:
+      - type: uint8
```
yuslepukhin: >type: uint8 [](start = 8, length = 11)

I think we support a full spectre of types.

src/Tools/CodeGenerator/Featurizers.yaml
wangyems: yes

src/Tools/CodeGenerator/Featurizers.yaml
wangyems: I don't think so as this is a parameter that passing in from constructor

yuslepukhin: Are you referring to a State to deserialze? Where do get that state from? This does not have any training part.

---
In reply to: [388535793](https://github.com/microsoft/FeaturizersLibrary/pull/159#discussion_r388535793) [](ancestors = 388535793)

davidbrownellWork: This is state that is created by the data scientist when constructing the model. It is not learned from the data, but rather a value that they provide.

A value of uint8 gives us a MaxHorizon value of 255; Is that enough for all scenarios? I think that the answer is probably yes, but it might be good to bump this up to a uint16 just to make sure.

davidbrownellWork (changes requested):

src/Tools/CodeGenerator/Featurizers.yaml
wangyems: yeah this is only a transformer. what I mean is that(I am not very clear how ort does this, but I guess maxHorizon still serves as parameters, probably in ctx) maxHorizon is highly possible a limited and small number, do we really need to support full spectre for that?

wangyems: > This is state that is created by the data scientist when constructing the model. It is not learned from the data, but rather a value that they provide.
> 
> A value of uint8 gives us a MaxHorizon value of 255; Is that enough for all scenarios? I think that the answer is probably yes, but it might be good to bump this up to a uint16 just to make sure.

sure

yuslepukhin: This would come from the attribute. I think it would be better if we just use more natural types for the CPU such as uint43 or int64

---
In reply to: [388542380](https://github.com/microsoft/FeaturizersLibrary/pull/159#discussion_r388542380) [](ancestors = 388542380)

justinormont: From the AutoML perspective, UINT32 should suffice.

davidbrownellWork (dismissed):

src/Tools/CodeGenerator/Featurizers.yaml
wangyems: looks like uint32 is better

justinormont (approved):

src/Tools/CodeGenerator/Featurizers.yaml
```diff
@@ -639,6 +639,75 @@ featurizers:
...
+          +------------+-------+-------+----------------+
+          | 2017-03-01 | A     | 2     | 2              |
+          +------------+-------+-------+----------------+
+          | 2017-04-01 | A     | 3     | 1              |
+          +------------+-------+-------+----------------+
+          | 2017-04-01 | A     | 3     | 2              |
+          +------------+-------+-------+----------------+
+
+    configuration_params:
+      - type: uint32
```
davidbrownellWork: By changing this to uint32, it implies that we can have a horizon of 2^32. Is this really what we want? My preference is:

1) uint16
2) uint8
3) uint32

davidbrownellWork (changes requested):

src/Tools/CodeGenerator/Featurizers.yaml
yuslepukhin: Can you explain why is this your preference? It really can not be a uint8 as we need to use it in arithmetic operations such as a counter. We'd need to cast it to one of the types.  Why not use the natural datatype for which the processor has built in instructions?

It will also cause padding in the class anyway.

davidbrownellWork: This is the type that will be expressed in the estimator's constructor as the number of rolling windows to calculate for each "row". When we restrict the type to something smaller, we are communicating to the caller the range of acceptable values. I look at it as a form of documentation to the caller.

In the internal Estimator implementation, we can convert this to whatever type is appropriate based on how it is used (for example, the class might store this value as uint32 even though the incoming type is something smaller).

justinormont: We have a bit of docs about the max horizon parameter: https://docs.microsoft.com/en-us/azure/machine-learning/how-to-configure-auto-train

> `max_horizon`: Defines the length of time you want to predict out based on the periodicity of the training data. For example if you have training data with daily time grains, you define how far out in days you want the model to train for.

 While we expect most users would be ok with UINT8, a subset of users have larger needs.

Let's assume your interval of data is one day. A UINT8 would let you predict 255 days in the future. It seems reasonable for someone to ask, "how many oranges will I sell for the next year", which a UINT8 would not allow.

A reasonable upper limit for this is also beyond a UINT16 for similar reasons. Let's assume your data arrives every second. A UINT16 would let you predict 65,535 seconds in the future (18 hours). It also seems reasonable for someone to ask, "what will my traffic be for the next day", which a UINT16 would not allow.

A UINT32 with data arriving every second, would be 136 years. Certainly more than we expect users to request.

Hence my recommendation for a UINT32 to store the max horizon.

davidbrownellWork: Thanks, Justin - I'm convinced!

justinormont: Correct, this is a multiplier of the period of the user's data, and it limits how far into the future the user can predict to `<= period * max_horizon`.

While most users would do ok with only a UINT8, some may need larger than a UINT16; hence the UINT32 is best.

davidbrownellWork: That's fine, we can make it uint32. 

However, can you go into a bit more detail about scenarios when an uint32 would be necessary? I want to make sure that we have a good understanding of those as we are talking about a large enough data set that it might alter other designs. Thanks, Justin.

PR #160, opened 2020-03-05
Jin/minimumfrequency

angryjinyan:
Wrapped the FrequencyEstimator by GrainEstimator to get time frequency for each grain.

Created the GlobalMinimumFrequencyEstimator to get the minimum time frequency across grains.


src/Featurizers/Components/FrequencyEstimator.h
```diff
@@ -0,0 +1,169 @@
...
+    }
+    FrequencyType currFrequency(input - _lastObserved);
+    if (currFrequency <= _frequency) {
+        _frequency = currFrequency;
+    }
+    _lastObserved = input;
+}
+
+FrequencyAnnotation Details::FrequencyTrainingOnlyPolicy::complete_training(void) {
+    return FrequencyAnnotation(std::move(_frequency));
```
davidbrownellWork: >return FrequencyAnnotation(std::move(_frequency)); [](start = 4, length = 50)

Should it be an error to complete training without having seen a value?

src/Featurizers/Components/FrequencyEstimator.h
```diff
@@ -0,0 +1,169 @@
...
+#include <limits>
+
+#include "TrainingOnlyEstimatorImpl.h"
+
+namespace Microsoft {
+namespace Featurizer {
+namespace Featurizers {
+namespace Components {
+
+// TODO: This should be removed in favor of the StatisticalMetricsEstimator
```
davidbrownellWork: I don't understand this comment. Is this saying that frequency detection should become part of the StatisticalMetricsEstimator?

src/Featurizers/Components/GlobalMinimumFrequencyEstimator.h
```diff
@@ -0,0 +1,222 @@
...
+namespace Featurizer {
+namespace Featurizers {
+namespace Components {
+
+// TODO: This should be removed in favor of the StatisticalMetricsEstimator
+
+static constexpr char const * const         GlobalMinimumFrequencyEstimatorName("GlobalMinimumFrequencyEstimator");
+static constexpr char const * const         GrainedFrequencyEstimatorName("GrainedFrequencyEstimator");
+
+using GrainedFrequencyEstimator = GrainEstimatorImpl<std::string, FrequencyEstimator<>>;
```
davidbrownellWork: >std::string [](start = 53, length = 11)

To fit with existing conventions, grain should be represented as `std::vector<std::string>`

src/Featurizers/Components/FrequencyEstimator.h
angryjinyan: No. It's a comment I forgot to remove.

src/Featurizers/Components/GlobalMinimumFrequencyEstimator.h
davidbrownellWork: >using GrainedFrequencyEstimator [](start = 0, length = 31)

This using statement should be a template with the MaxNumTrainingItemsV set to std::numeric_limits<size_t>::max() as a default. That template value should be provided to FrequencyEstimator.

src/Featurizers/Components/GlobalMinimumFrequencyEstimator.h
```diff
@@ -0,0 +1,222 @@
...
+    FEATURIZER_MOVE_CONSTRUCTOR_ONLY(GlobalMinimumFrequencyAnnotation);
+};
+
+namespace Details {
+
+/////////////////////////////////////////////////////////////////////////
+///  \class         GlobalMinimumFrequencyTrainingOnlyPolicy
+///  \brief         `GlobalMinimumFrequencyEstimator` implementation details.
+///
+template <typename EstimatorT>
```
davidbrownellWork: I don't think that this needs to be template, does it?

src/Featurizers/Components/FrequencyEstimator.h
```diff
@@ -0,0 +1,169 @@
...
+    // |
+    // |  Private Data
+    // |
+    // ----------------------------------------------------------------------
+    FrequencyType _frequency;
+    InputType _lastObserved;
+
+    bool _hasSeenValues;
+};
+}
```
michaelgsharp: Namespace comment here.

src/Featurizers/Components/GlobalMinimumFrequencyEstimator.h
```diff
@@ -0,0 +1,222 @@
...
+};
+}
+/////////////////////////////////////////////////////////////////////////
+///  \typedef       GlobalMinimumFrequencyEstimator
+///  \brief         A training-only class that finds the min, max
+///
+template <
+    size_t MaxNumTrainingItemsV=std::numeric_limits<size_t>::max()
+>
+class GlobalMinimumFrequencyEstimator :
```
davidbrownellWork: >GlobalMinimumFrequencyEstimator [](start = 6, length = 31)

I didn't expect to see this implemented in this way - I thought that it would rely on annotations produced by the GrainedFrequencyEstimator rather that contain it. We have discussed how the LagLead/RollingWindow featurizer would work in relationship to the TimeSeriesImputer, but didn't expect to see those concepts extend to these components.

src/Featurizers/Components/FrequencyEstimator.h
michaelgsharp: Yes it should be. Jin, I fixed this in the original branch we started working on. Can you make sure to pull in the changes/tests from that branch?

src/Featurizers/Components/GlobalMinimumFrequencyEstimator.h
```diff
@@ -0,0 +1,222 @@
...
+
+template <typename EstimatorT>
+GlobalMinimumFrequencyAnnotation Details::GlobalMinimumFrequencyTrainingOnlyPolicy<EstimatorT>::complete_training(void) {
+    _estimator.complete_training();
+    GrainEstimatorAnnotation<std::string> const & data(GrainedFrequencyEstimator::get_annotation(_estimator.get_column_annotations(), 0, GrainedFrequencyEstimatorName));
+
+    GrainEstimatorAnnotation<std::string>::AnnotationMap annotation(data.Annotations);
+
+    FrequencyType freq(std::chrono::system_clock::duration::max().count());
+    for (auto it = annotation.begin(); it != annotation.end(); it++) {
```
davidbrownellWork: >for (auto it = annotation.begin(); it != annotation.end(); it++) { [](start = 4, length = 66)

You don't need to use iterators here, as you only care about the values.

`for(auto const & kvp : data.Annotations)`

src/Featurizers/Components/FrequencyEstimator.h
angryjinyan: good catch!

src/Featurizers/Components/GlobalMinimumFrequencyEstimator.h
angryjinyan: Implementation has changed to be like IndexMapEstimator. EstimatorT is needed to retrieve annotation.

src/Featurizers/Components/GlobalMinimumFrequencyEstimator.h
angryjinyan: Will resolve in offline discussion.

angryjinyan: Resolved. Change `std::unordered_map` to `std::map` to avoid compiler complaints about using `std::vector<std::string>>` as key type.

src/Featurizers/Components/GrainFeaturizerImpl.h
```diff
@@ -215,11 +213,9 @@ class GrainEstimatorImplBase : public BaseT {
     // |
     // ----------------------------------------------------------------------
     using EstimatorMap =
-        std::unordered_map<
+        std::map<
             GrainT,
-            EstimatorT,
-            std::hash<GrainT>,
```
wangyems: not a comment. 
if using unordered_map, hash function should be customly defined, std::hash<T> probably not gonna work

src/Featurizers/Components/GlobalMinimumFrequencyEstimator.h
```diff
@@ -0,0 +1,200 @@
...
+template <typename EstimatorT>
+class GlobalMinimumFrequencyTrainingOnlyPolicy {
+public:
+
+    // ----------------------------------------------------------------------
+    // |
+    // |  Public Types
+    // |
+    // ----------------------------------------------------------------------
+    using InputType                     = std::tuple<std::vector<std::string>, std::chrono::system_clock::time_point>;
```
michaelgsharp: In my ROllingWIndow I am defining the time_point as the first input in the tuple. Do you remember what you are doing in your laglead? we need to be consistent. Also, there will be other inputs in the featurizers that follow this. Do we need to account for them here?

src/Featurizers/Components/GrainFeaturizerImpl.h
yuslepukhin: You just need to define a specialization for your type for std::hash<>, that is how  it is used.

---
In reply to: [389183880](https://github.com/microsoft/FeaturizersLibrary/pull/160#discussion_r389183880) [](ancestors = 389183880)

src/Featurizers/Components/GlobalMinimumFrequencyEstimator.h
yuslepukhin: Just wondering, if you declare it is a tuple with a vector, then you need to copy all that data to populate the vector to put it in the tuple. If you continue down this road, neither ORT nor C++ will make ML.NET algorithms fast. Can you make input a structure with iterators that point to the data? We already did that.

---
In reply to: [389184166](https://github.com/microsoft/FeaturizersLibrary/pull/160#discussion_r389184166) [](ancestors = 389184166)

src/Featurizers/Components/FrequencyEstimator.h
davidbrownellWork: This still needs to be addressed.

---
In reply to: [388728448](https://github.com/microsoft/FeaturizersLibrary/pull/160#discussion_r388728448) [](ancestors = 388728448)

src/Featurizers/Components/GlobalMinimumFrequencyEstimator.h
```diff
@@ -0,0 +1,200 @@
...
+namespace Microsoft {
+namespace Featurizer {
+namespace Featurizers {
+namespace Components {
+
+static constexpr char const * const         GlobalMinimumFrequencyEstimatorName("GlobalMinimumFrequencyEstimator");
+static constexpr char const * const         GrainedFrequencyEstimatorName("GrainedFrequencyEstimator");
+
+template <size_t MaxNumTrainingItemsV=std::numeric_limits<size_t>::max()>
+using GrainedFrequencyEstimator = GrainEstimatorImpl<std::vector<std::string>, FrequencyEstimator<>>;
```
davidbrownellWork: >FrequencyEstimator<> [](start = 79, length = 20)

`FrequencyEstiamtor<MaxNumTrrainingItemsV>`

davidbrownellWork: These should be defined in FrequencyEstimator.h.

src/Featurizers/Components/GlobalMinimumFrequencyEstimator.h
```diff
@@ -0,0 +1,200 @@
...
+    FEATURIZER_MOVE_CONSTRUCTOR_ONLY(GlobalMinimumFrequencyAnnotation);
+};
+
+namespace Details {
+
+/////////////////////////////////////////////////////////////////////////
+///  \class         GlobalMinimumFrequencyTrainingOnlyPolicy
+///  \brief         `GlobalMinimumFrequencyEstimator` implementation details.
+///
+template <typename EstimatorT>
```
davidbrownellWork: >EstimatorT [](start = 19, length = 10)

Rename to `GlobalMinimumFrequencyEstimatorT`, as this can't be just any type.

src/Featurizers/Components/GlobalMinimumFrequencyEstimator.h
davidbrownellWork: Using `std::vector<std::string>` to represent a grain is a hack introduced with TimeSeriesImputer that is now breaking down. We need to address this before the next release, but is something that we can move forward with now knowing that it has to be addressed in the near future.

---
In reply to: [389185922](https://github.com/microsoft/FeaturizersLibrary/pull/160#discussion_r389185922) [](ancestors = 389185922,389184166)

src/Featurizers/Components/GlobalMinimumFrequencyEstimator.h
```diff
@@ -0,0 +1,200 @@
...
+void Details::GlobalMinimumFrequencyTrainingOnlyPolicy<EstimatorT>::fit(InputType const &) {
+    throw std::runtime_error("This will never be called");
+}
+
+template <typename EstimatorT>
+GlobalMinimumFrequencyAnnotation Details::GlobalMinimumFrequencyTrainingOnlyPolicy<EstimatorT>::complete_training(void) {
+    EstimatorT                                         const & estimator(static_cast<EstimatorT const &>(*this));
+    GrainEstimatorAnnotation<std::vector<std::string>> const & data(GrainedFrequencyEstimator<>::get_annotation(estimator.get_column_annotations(), estimator.get_column_index(), GrainedFrequencyEstimatorName));
+
+    GrainEstimatorAnnotation<std::vector<std::string>>::AnnotationMap annotation(data.Annotations);
```
davidbrownellWork: >GrainEstimatorAnnotation<std::vector<std::string>>::AnnotationMap [](start = 4, length = 65)

const &

src/Featurizers/Components/FrequencyEstimator.h
```diff
@@ -0,0 +1,167 @@
...
+// |  Details::FrequencyTrainingOnlyPolicy
+// |
+// ----------------------------------------------------------------------
+Details::FrequencyTrainingOnlyPolicy::FrequencyTrainingOnlyPolicy(void) :
+    _frequency(std::chrono::system_clock::duration::max().count()),
+    _hasSeenValues(false) {
+}
+
+void Details::FrequencyTrainingOnlyPolicy::fit(InputType const &input) {
+    if (!_hasSeenValues) {
```
michaelgsharp: Switch to ntd::optional.
Refactor to only one return statement.
Add a const reference after the value has been set.

src/Featurizers/Components/GlobalMinimumFrequencyEstimator.h
michaelgsharp: Making a copy here. Use reference instead.

src/Featurizers/Components/GlobalMinimumFrequencyEstimator.h
davidbrownellWork: >std::tuple<std::vector<std::string>, std::chrono::system_clock::time_point>; [](start = 42, length = 76)

`std::tuple<std::vector<std::string> const &, std::chrono::system_clock::time_point const &>`

PR #161, opened 2020-03-09
Refactor for grains

davidbrownellWork:


src/Featurizers/Components/GrainFeaturizerImpl.h
```diff
@@ -499,19 +501,24 @@ Impl::GrainEstimatorImplBase<BaseT, GrainT, EstimatorT, MaxNumTrainingItemsV>::G
...
-        )
+            // Create the name of the GrainEstimator as "Grain<EstimatorName>". This is slightly more
+            // complicated than what would normally be expected, as we have to create an instance
+            // of the estimator before we can get its name.
+
+            EstimatorT const                estimator(createFunc(std::move(pAllColumnAnnotations)));
+
+            return std::string("Grain") + estimator.Name;
+        }(),
+        pAllColumnAnnotations
```
michaelgsharp: Do we need to worry about moving `pAllColumnAnnotations` on line 514 and then using it again on line 518?

davidbrownellWork: In this case, we don't as the lambda makes a copy of the value in its reference list on line 506.

PR #162, opened 2020-03-10
Jin/minimumfrequency

michaelgsharp:
Original PR #160 

Wrapped the FrequencyEstimator by GrainEstimator to get time frequency for each grain.

Created the GlobalMinimumFrequencyEstimator to get the minimum time frequency across grains.


src/Featurizers/Components/FrequencyEstimator.h
```diff
@@ -0,0 +1,169 @@
...
+#include "TrainingOnlyEstimatorImpl.h"
+#include "GrainFeaturizerImpl.h"
+
+namespace Microsoft {
+namespace Featurizer {
+namespace Featurizers {
+namespace Components {
+
+static constexpr char const * const         FrequencyEstimatorName("FrequencyEstimator");
+static constexpr char const * const         GrainFrequencyEstimatorName("GrainFrequencyEstimator");
```
davidbrownellWork: I don't believe that this is necessary anymore.

src/Featurizers/Components/GlobalMinimumFrequencyEstimator.h
```diff
@@ -0,0 +1,195 @@
...
+    // |
+    // ----------------------------------------------------------------------
+    static constexpr char const * const     NameValue = GlobalMinimumFrequencyEstimatorName;
+
+    // ----------------------------------------------------------------------
+    // |
+    // |  Public Methods
+    // |
+    // ----------------------------------------------------------------------
+    GlobalMinimumFrequencyTrainingOnlyPolicy(void);
```
davidbrownellWork: >GlobalMinimumFrequencyTrainingOnlyPolicy(void); [](start = 4, length = 47)

It doesn't look like this constructor is doing anything and it can be removed.

src/Featurizers/Components/FrequencyEstimator.h
michaelgsharp: We are using it on line 178 in the `GlobalMinimumFrequencyEstimator.h` to get the annotations. We could just use it there and not define it, but we were thinking that if we ever want to get these annotations anywhere else that it would make sense to have it defined somewhere.

src/Featurizers/Components/GlobalMinimumFrequencyEstimator.h
michaelgsharp: Good catch. I have removed it now.

src/Featurizers/Components/FrequencyEstimator.h
michaelgsharp: Per offline discussion, merging for now and will create a new PR with these changes.

PR #163, opened 2020-03-11
Yaml update for LagLeadOperatorFeaturizer

angryjinyan:
Update yaml file for LagLeadOperatorFeaturizer. Since one can use this featurizer for multiple lag or lead orders and NaN is returned for not found data, vector of double is the return type.

Type for start_offset(horizon) is set to uint32 and type for lag or lead orders is set to int32(negative value means lead) for now. Subject to changes for future discussions.

### DIFF
--- a/src/Tools/CodeGenerator/Featurizers.yaml
+++ b/src/Tools/CodeGenerator/Featurizers.yaml
@@ -1416,9 +1416,17 @@ featurizers:
     description: |-
       Copying values from back or in the future per grain. Works for general time series data sets.
 
-      Offsets are the lag or lead orders we use to find the row in the past or in the future. Positive offset means looking into the future and negative offset means looking back in this featurizer.
-      
-      Horizon is an unsigned int to mark the starting position to look for the data, ranging from 1 to N.
+      The Horizon represents the maximum value in a range [1, N], where each element in that range is a delta applied to each offset. The resulting matrix will be in the form:
+
+      [
+      [value[1 + offset[K - 1]], value[2 + offset[K - 1]], ..., value[N + offset[K - 1]],
+      [value[1 + offset[K - 2]], value[2 + offset[K - 2]], ..., value[N + offset[K - 2]],
+      ...
+      [value[1 + offset[1]], value[2 + offset[1]], ..., value[N + offset[1]],
+      [value[1 + offset[0]], value[2 + offset[0]], ..., value[N + offset[0]]
+      ]
+
+      The resulting matrix size is K rows x N cols, where K is the number of offsets and N is the horizon.
 
       Horizon and offsets should be passed in during construction. Offsets are passed in as a vector of ints so multiple lag orders can be applied within one featurizer call. 
       Output type is a matrix of double where rows are grouped by different offsets and columns are grouped by horizon.
--- a/src/Tools/CodeGenerator/Featurizers.yaml
+++ b/src/Tools/CodeGenerator/Featurizers.yaml
@@ -1418,54 +1418,47 @@ featurizers:
       For sparse time series data sets, forward fill imputer will be applied so input time series would always be continuous
       and have a consistent time interval between rows.
 
-      Input type for this featurizer is a tuple of time point of current row, the grain, the delta offset and target column to lag or lead.
+      Input type for this featurizer is a tuple of time point of current row, the grain and target column to lag or lead.
 
-      We use "delta offset" to refer max horizon because it should apply to all offset computations to find the data we want.
+      We use "delta offset" to refer max horizon because it should apply to all offset computations to find the data we want. Delta offset and lag or lead offsets should
+      be passed in during construction.
+      Lag offsets are passed in as a vector of ints so multiple lag orders can be applied within one featurizer call. Positive lag offset means looking into the future and negative lag offset means looking back in this featurizer.
+      Output type is matrix of double where rows are grouped by different lag or lead orders and columns are grouped by delta offset.
 
-      Lag offset is passed in as a vector of ints so multiple lag orders can be applied within one featurizer call. Negative lag offset means lead in this featurizer.
 
       C++-style pseudo signature:
-        template <typename T> vector<double> execute(std::tuple<std::chrono::system_clock::time_point const &, std::vector<std::string> const &, uint32 const &, T const &> const &value);
+        template <typename T> matrix<double> execute(std::tuple<std::chrono::system_clock::time_point const &, std::vector<std::string> const &, T const &> const &value);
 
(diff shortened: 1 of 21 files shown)


src/Tools/CodeGenerator/Featurizers.yaml
```diff
@@ -1418,13 +1418,12 @@ featurizers:
...
-      Input type for this featurizer is a tuple of time point of current row, target column to lag or lead
-      and the horizon origin for the row.
+      Input type for this featurizer is a tuple of time point of current row, the grain column, the horizon and target column to lag or lead.
 
-      Minus lags means lead in this featurizer.
+      We use "start offset" to refer max horizon because it marks the position where the data we want starts and minus lags for lead in this featurizer.
 
       C++-style pseudo signature:
-        template <typename T> vector<double> execute(std::tuple<std::chrono::system_clock::time_point, T, std::chrono::system_clock::time_point> const &value);
+        template <typename T> vector<double> execute(std::tuple<std::chrono::system_clock::time_point, std::vector<std::string>, uint32, T> const &value);
```
davidbrownellWork: >std::chrono::system_clock::time_point, std::vector<std::string>, uint32, T [](start = 64, length = 74)

Each element of the tuple should be `const &`

davidbrownellWork:
>           |time(YYYY-MM-DD)| target| horizon_origin   | target_lag_1      |

Update this table to match the function definition

---
Refers to: src/Tools/CodeGenerator/Featurizers.yaml:1431 in 0b351d6. [](commit_id = 0b351d6b393896cf9755911736737699666d938a, deletion_comment = False)

src/Tools/CodeGenerator/Featurizers.yaml
```diff
@@ -1481,17 +1473,19 @@ featurizers:
           - float
           - double
 
+    # Output type should be changed to matrix<nonstd::optional<T>> and input should support all types
+    # after yaml file support nonstd::optional type
```
davidbrownellWork: Add a TODO prefix so that this comment appears in searches for "TODO"

angryjinyan: Good point. Added TODO prefix.

src/Tools/CodeGenerator/Featurizers.yaml
```diff
@@ -1414,58 +1414,50 @@ featurizers:
...
-      Computes lags and leads for values. Works for general time series data sets.
-      For sparse time series data sets, forward fill imputer will be applied so input time series would always be continuous
-      and have a consistent time interval between rows.
+      Copying values from back or in the future per grain. Works for general time series data sets.
 
-      Input type for this featurizer is a tuple of time point of current row, target column to lag or lead
-      and the horizon origin for the row.
+      Offsets are the lag or lead orders we use to find the row in the past or in the future. Positive offset means looking into the future and negative offset means looking back in this featurizer.
+      
+      Horizon is an unsigned int to mark the starting position to look for the data, ranging from 1 to N.
```
davidbrownellWork: The Horizon represents the maximum value in a range [1, N), where each element in that range is a delta applied to each offset. The resulting matrix will be in the form:

[ 
  [value[1 + offset[0]], value[2 + offset[0]], ..., value[N -1 + offset[0]],
  [value[1 + offset[1]], value[2 + offset[1]], ..., value[N -1 + offset[1]],
  ...
  [value[1 + offset[K - 1]], value[2 + offset[K - 1]], ..., value[N-1 + offset[K - 1]]
]

The resulting matrix size is K rows x N cols, where K is the number of offsets and N is the horizon.

(I think that I have this backwards)

src/Tools/CodeGenerator/Featurizers.yaml
```diff
@@ -1414,58 +1414,57 @@ featurizers:
...
-      and have a consistent time interval between rows.
+      Copying values from back or in the future per grain. Works for general time series data sets.
 
-      Input type for this featurizer is a tuple of time point of current row, target column to lag or lead
-      and the horizon origin for the row.
+      The Horizon represents the maximum value in a range [1, N], where each element in that range is a delta applied to each offset. The resulting matrix will be in the form:
 
-      Minus lags means lead in this featurizer.
+      [
+      [value[offset[0] - N], value[offset[0] - (N - 1)], ..., value[offset[0] - 1],
```
davidbrownellWork: >[value[offset[0] - N], value[offset[0] - (N - 1)], ..., value[offset[0] - 1], [](start = 6, length = 77)

Missing trailing ']'

angryjinyan: Good catch!

src/Tools/CodeGenerator/Featurizers.yaml
```diff
@@ -1414,58 +1414,57 @@ featurizers:
     num_output_columns: 1
     has_dynamic_output: true
     description: |-
-      Computes lags and leads for values. Works for general time series data sets.
-      For sparse time series data sets, forward fill imputer will be applied so input time series would always be continuous
-      and have a consistent time interval between rows.
+      Copying values from back or in the future per grain. Works for general time series data sets.
```
michaelgsharp: Now that we are removing the time information, this will work with any sorted dataset and not just timeseries datasets. I would also change the language from "back" to something else if possible. I think "back" isn't very clear. Maybe "past"? Or "prior"? What do you think?

src/Tools/CodeGenerator/Featurizers.yaml
```diff
@@ -1481,17 +1480,18 @@ featurizers:
           - float
           - double
 
+    # TODO: Output type should be changed to matrix<nonstd::optional<T>> and input should support all types after yaml file support nonstd::optional type
```
michaelgsharp: Isn't this always going to return a type double and not T? I thought that they said that this doesn't make sense for non-numeric values? Or are we just going to change to T so that we can support it even if its not something they use?

src/Tools/CodeGenerator/Featurizers.yaml
davidbrownellWork: I like "prior".

src/Tools/CodeGenerator/Featurizers.yaml
davidbrownellWork: Given that it is just copying values, it seems reasonable that it could copy any type and not just doubles.

src/Tools/CodeGenerator/Featurizers.yaml
angryjinyan: Makes sense. Changed to "prior".

PR #164, opened 2020-03-12
Added FilterEstimatorImpl.h and tests

davidbrownellWork:


src/Featurizers/Components/FilterFeaturizerImpl.h
```diff
@@ -0,0 +1,641 @@
...
+    void flush_impl(typename BaseType::CallbackFunction const &/*callback*/) override {
+        _pTransformer->flush(
+            [/*&callback*/](typename ChainedTransformerT::TransformedType /*value*/) {
+                // If errors are raised here, it is because the ChainedTransformer has dynamic output.
+                // This functionality isn't implemented right now, as we aren't sure of a good
+                // way to augment `value` with the row information provided when the corresponding
+                // call to `execute` was made. Furthermore, we don't have a good way to associated
+                // calls to this callback with multiple calls to `execute`. Further design changes
+                // will be necessary if this functionality is required in the future.
+                throw std::runtime_error("Transformers that rely on flush can't be wrapped in this object at this time");
```
michaelgsharp: I think they may be an issue earlier than I was thinking. When we first talked about this I was thinking we would only have to wrap 1 step of the pipeline. After our conversations I realize that if a "normal" featurizer that is wrapped in the filter featurizer returns output, then the rest of the featurizers in the pipeline will need to be filtered as well. This means that if an earlier step is not dynamic, but a final step is, that we need to include this functionality.

Fortunately, in this current case the featurizer that is getting wrapped only produces an annotation and not any output, so it is still not an issue. I just wanted to call it out.

src/Featurizers/Components/UnitTests/FilterFeaturizerImpl_UnitTest.cpp
```diff
@@ -0,0 +1,552 @@
...
+        callback(value + _delta);
+    }
+};
+
+TEST_CASE("SimpleSumTransformer") {
+    // ----------------------------------------------------------------------
+    using FilterTransformer =
+        Components::FilterTransformer<
+            SimpleSumTransformer,
+            std::tuple<char, int>,
```
michaelgsharp: Does it make a difference that these aren't const refs?

src/Featurizers/Components/UnitTests/FilterFeaturizerImpl_UnitTest.cpp
```diff
@@ -0,0 +1,552 @@
...
+TEST_CASE("TupleOutTransformer") {
+    // ----------------------------------------------------------------------
+    using FilterTransformer =
+        Components::FilterTransformer<
+            TupleOutTransformer,
+            std::tuple<int, char>,
+            0
+        >;
+
+    using Results                           = std::vector<std::tuple<int, char, int, int>>;
```
michaelgsharp: If I understood the original test transformer, it takes in a value and outputs a tuple. It looks like the filter featurizer is then unpacking that tuple for a final result `std::tuple<int, char, int, int>`. Is there ever a situation where we woudln't want that and we would want the result to be `std::tuple<int, char, std::tuple<int, int>>`? I can't think of any off of the top of my head, but it is something I thought about.

davidbrownellWork: We could go this route, but then we would run into downstream composability problems. For example, it would be difficult to Filter with the second element of `tuple<int, int>` via the third element of `tuple<int, char, tuple<int, int>>`.

src/Featurizers/Components/UnitTests/FilterFeaturizerImpl_UnitTest.cpp
davidbrownellWork: I realized this as I was writing tests and got lazy. I think that I should add code to verify that all input elements are const references and update tests, but I think that I will do that as a separate PR.

PR #165, opened 2020-03-12
Small fixes not in the original PR

michaelgsharp:
Removed unused constructor.

Moved grain featurizer name.

### DIFF
--- a/src/Featurizers/Components/FrequencyEstimator.h
+++ b/src/Featurizers/Components/FrequencyEstimator.h
@@ -15,7 +15,6 @@ namespace Featurizers {
 namespace Components {
 
 static constexpr char const * const         FrequencyEstimatorName("FrequencyEstimator");
-static constexpr char const * const         GrainFrequencyEstimatorName("GrainFrequencyEstimator");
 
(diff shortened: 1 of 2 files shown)


src/Featurizers/Components/FrequencyEstimator.h
```diff
@@ -108,6 +107,8 @@ template <
 >
 using FrequencyEstimator                       = TrainingOnlyEstimatorImpl<Details::FrequencyTrainingOnlyPolicy, MaxNumTrainingItemsV>;
 
+// Normally we want the featurizer name at the top of the file, but GrainFeaturizers are just type defs, so we want the name next to it.
+static constexpr char const * const         GrainFrequencyEstimatorName("GrainFrequencyEstimator");
```
michaelgsharp: I moved this down, but would I be better to move the name and the typedef up instead? Just a thought.

davidbrownellWork: This looks good to me.

---
In reply to: [391819433](https://github.com/microsoft/FeaturizersLibrary/pull/165#discussion_r391819433) [](ancestors = 391819433)

PR #167, opened 2020-03-13
Initial tuple elements passed to FilterFeaturizerImpl must be references

davidbrownellWork:


src/Featurizers/Components/FilterFeaturizerImpl.h
```diff
@@ -32,6 +32,16 @@ struct FilterFeaturizerTraits {
     // ----------------------------------------------------------------------
     static_assert(Details::IsTuple<InputTupleT>::value, "'InputTupleT' must be a tuple");
 
+    // The first N elements in the tuple represent the original input data and must therefore be
+    // references. However, not all of the tuple elements must be references. For example, if this
+    // FilterFeaturizer is part of a pipeline with upstream FilterFeaturizers, the first elements
+    // associated with the original input will be references, but the latter elements will be the
+    // result of the transformer (and not references).
```
michaelgsharp: Why aren't the results also references? Will that not mean that those values are copied for the rest of the pipeline?

davidbrownellWork: The elements can be moved (which would prevent the copy). They can't be references as we can't make assumptions about how the results were created (at least, I haven't reasoned about the assumptions that can or can't be made).

PR #168, opened 2020-03-14
Added FilterSinkFeaturizer.h and tests

davidbrownellWork:


src/Featurizers/Components/UnitTests/FilterDecoratorFeaturizer_UnitTest.cpp
```diff
@@ -0,0 +1,117 @@
...
+#include <ostream>
+#include <tuple>
+
+template <size_t IndexV, typename...Ts>
+void PrintTupleElement(std::ostream &os, std::tuple<Ts...> const &value, typename std::enable_if<IndexV + 1 == sizeof...(Ts)>::type * =nullptr) {
+    os << std::get<IndexV>(value);
+}
+
+template <size_t IndexV, typename... Ts>
+void PrintTupleElement(std::ostream &os, std::tuple<Ts...> const &value, typename std::enable_if<IndexV + 1 != sizeof...(Ts)>::type * =nullptr) {
```
wangyems: I remember Traits supports ToString() of tuple type

davidbrownellWork: Good catch! I have a change queued up after this one to fix another issue with the Filter estimators and will add this change to that branch before I create the PR.

PR #171, opened 2020-03-18
LagLeadOperatorFeaturizer

angryjinyan:
Finished the transformer logic. Need to add GrainEstimator wrapper to LagLeadOperatorEstimator


src/Featurizers/Components/WindowFeaturizerBase.h
```diff
@@ -135,6 +135,8 @@ namespace Components {
         // ----------------------------------------------------------------------
         CircularBuffer(size_t max_size);
 
+        size_t size();
```
yuslepukhin: >e(); [](start = 18, length = 4)

const ?

src/Featurizers/Components/WindowFeaturizerBase.h
```diff
@@ -293,6 +295,11 @@ CircularBuffer<T>::CircularBuffer(size_t max_size) : _max_size(max_size), _start
     _data.reserve(_max_size);
 }
 
+template <class T>
+size_t CircularBuffer<T>::size() {
```
yuslepukhin: >() { [](start = 30, length = 4)

const

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,234 @@
...
+
+namespace Microsoft {
+namespace Featurizer {
+namespace Featurizers {
+
+/////////////////////////////////////////////////////////////////////////
+///  \class         LagLeadOperatorTransformer
+///  \brief         Copy values from prior data or future data
+///
+template <typename InputTupleT>
```
yuslepukhin: >InputTupleT [](start = 19, length = 11)

Nice!

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,234 @@
...
+    // ----------------------------------------------------------------------
+    // |
+    // |  Private Data
+    // |
+    // ----------------------------------------------------------------------
+    Components::CircularBuffer<TargetType>  _buffer;
+    std::int64_t const                      _max_future_offset;
+    std::int64_t const                      _max_prior_offset;
+    std::uint32_t const                     _horizon;
+    std::vector<std::int64_t> const         _offsets;
```
yuslepukhin: >std::vector<std::int64_t> const   [](start = 4, length = 33)

This can not be moved if it is const and so the fact that is copy constructor is prohibited becomes irrelevant, everything is being copied.

angryjinyan: This vector is only copied in the constructor so it may not cost that much. Do you have any suggestions on this?

src/Featurizers/LagLeadOperatorFeaturizer.h
angryjinyan: Good call. Descriptions added.

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,234 @@
...
+            ++_index;
+            ++_numPending;
+            _buffer.push(std::get<1>(input));
+            return;
+        }
+        Microsoft::Featurizer::ColMajMatrix<std::double_t> ret(_offsets.size(), _horizon);
+        for (std::uint32_t col = 0; col < _horizon; ++col) {
+            for (size_t row = 0; row < _offsets.size(); ++row) {
+                if ((_index - _numPending - (_horizon - 1 - col) + _offsets[row] < 0)) {
+                    ret(static_cast<long long>(row), col) = std::nan("");
```
yuslepukhin: >"" [](start = 69, length = 2)

What is the meaning of this, passing an empty const char* pointer to a function that expects integral types?

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,234 @@
...
+
+    // MSVC runs into problems when the declaration and definition are separated
+    void execute_impl(typename BaseType::InputType const &input, typename BaseType::CallbackFunction const &callback) override {
+        if (_max_future_offset != 0 && _index <= _max_future_offset + _horizon - 1) {
+            ++_index;
+            ++_numPending;
+            _buffer.push(std::get<1>(input));
+            return;
+        }
+        Microsoft::Featurizer::ColMajMatrix<std::double_t> ret(_offsets.size(), _horizon);
```
yuslepukhin: >ColMajMatrix [](start = 31, length = 12)

Onnx runtime operates in RowMajor matrices. This means it will have to convert it. Do you need a column major in this case?

src/Featurizers/LagLeadOperatorFeaturizer.h
angryjinyan: It's creating a quiet NaN and the method requires a const char * parameter which I think I can leave as empty in this case.

src/Featurizers/LagLeadOperatorFeaturizer.h
angryjinyan: No I don't. Have changed to RowMajor matrices.

src/Featurizers/LagLeadOperatorFeaturizer.h
yuslepukhin: I suggest phibiting both copy and move for the transformer

---
In reply to: [394597621](https://github.com/microsoft/FeaturizersLibrary/pull/171#discussion_r394597621) [](ancestors = 394597621)

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,235 @@
...
+        if (_max_future_offset != 0 && _index <= _max_future_offset + _horizon - 1) {
+            ++_index;
+            ++_numPending;
+            _buffer.push(std::get<1>(input));
+            return;
+        }
+        Microsoft::Featurizer::RowMajMatrix<std::double_t> ret(_offsets.size(), _horizon);
+        for (std::uint32_t col = 0; col < _horizon; ++col) {
+            for (size_t row = 0; row < _offsets.size(); ++row) {
+                if ((_index - _numPending - (_horizon - 1 - col) + _offsets[row] < 0)) {
```
yuslepukhin: >offsets[row] [](start = 68, length = 12)

How do we make sure that row is not out of bounds?

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,235 @@
...
+    // MSVC runs into problems when the declaration and definition are separated
+    void execute_impl(typename BaseType::InputType const &input, typename BaseType::CallbackFunction const &callback) override {
+        if (_max_future_offset != 0 && _index <= _max_future_offset + _horizon - 1) {
+            ++_index;
+            ++_numPending;
+            _buffer.push(std::get<1>(input));
+            return;
+        }
+        Microsoft::Featurizer::RowMajMatrix<std::double_t> ret(_offsets.size(), _horizon);
+        for (std::uint32_t col = 0; col < _horizon; ++col) {
```
yuslepukhin: >std::uint32_t col = 0; col < _horizo [](start = 13, length = 36)

If the matrix is row major we should move cols into inner loop

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,235 @@
...
+            for (size_t row = 0; row < _offsets.size(); ++row) {
+                if ((_index - _numPending - (_horizon - 1 - col) + _offsets[row] < 0)) {
+                    ret(static_cast<long long>(row), col) = std::nan("");
+                }
+                else {
+                    ret(static_cast<long long>(row), col) = *std::get<0>(_buffer.range(
+                                                                                        1, 
+                                                                                        static_cast<size_t>(static_cast<long long>(_buffer.size() - _numPending - (_horizon - 1 - col)) + _offsets[row])
```
yuslepukhin: >_numPending - (_horizon - 1 - col) [](start = 148, length = 34)

Hard to read. So if horizon is 1 and col is 1 then the result is -1 which then add to the result, is it what you  want? Is it possible for the result to be negative. Bc if if you cast negative to size_t the result will be a very big positive integer.

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,235 @@
...
+                                                                                      ));
+                }
+            }
+        }
+        _buffer.push(std::get<1>(input));
+        ++_index;
+        callback(ret);
+    }
+
+    void flush_impl(typename BaseType::CallbackFunction const & callback) override {
```
yuslepukhin: >lush_impl( [](start = 10, length = 10)

Same comments apply here as the above function.

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,235 @@
...
+            ++_index;
+            ++_numPending;
+            _buffer.push(std::get<1>(input));
+            return;
+        }
+        Microsoft::Featurizer::RowMajMatrix<std::double_t> ret(_offsets.size(), _horizon);
+        for (std::uint32_t col = 0; col < _horizon; ++col) {
+            for (size_t row = 0; row < _offsets.size(); ++row) {
+                if ((_index - _numPending - (_horizon - 1 - col) + _offsets[row] < 0)) {
+                    ret(static_cast<long long>(row), col) = std::nan("");
```
yuslepukhin: >long long [](start = 36, length = 9)

Would it possible to use some fixed length type such as int64_t?

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,235 @@
...
+        static_cast<size_t>(std::abs(offsets[0])) : 
+        static_cast<size_t>(*std::max_element(offsets.cbegin(),offsets.cend()) - *std::min_element(offsets.cbegin(),offsets.cend()))))
+    ),
+    _max_future_offset(std::move(
+        *std::max_element(offsets.cbegin(), offsets.cend()) > 0 ?  *std::max_element(offsets.cbegin(), offsets.cend()) : 0
+    )),
+    _max_prior_offset(std::move(
+        *std::min_element(offsets.cbegin(), offsets.cend()) < 0 ?  -(*std::min_element(offsets.cbegin(), offsets.cend())) : 0
+    )),
+    _horizon(std::move(horizon)),
```
yuslepukhin: >std::move [](start = 13, length = 9)

What does this do?

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,235 @@
...
+                    }
+                }
+                callback(ret);
+                _numPending--;
+            }
+        }
+    }
+};
+
+/////////////////////////////////////////////////////////////////////////
```
yuslepukhin: I am not reviewing the estimator, perhaps someone else could do it.

src/Featurizers/LagLeadOperatorFeaturizer.h
angryjinyan: Col is also governed by horizon so (_horizon - 1 - col) will always be a positive number

angryjinyan: And this is a very good point. I just changed it to make it more readable.

src/Featurizers/LagLeadOperatorFeaturizer.h
angryjinyan: Good point! Changed to int64

src/Featurizers/Components/WindowFeaturizerBase.h
```diff
@@ -293,6 +295,11 @@ CircularBuffer<T>::CircularBuffer(size_t max_size) : _max_size(max_size), _start
     _data.reserve(_max_size);
 }
 
+template <class T>
+size_t CircularBuffer<T>::size() const {
+    return _data.size();
+}
+
```
michaelgsharp: I added this method too, so if my PR goes in first you should revert this file, and if yours goes in first ill revert mine. Easier to avoid a conflict that way.

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,237 @@
...
+namespace Microsoft {
+namespace Featurizer {
+namespace Featurizers {
+
+/////////////////////////////////////////////////////////////////////////
+///  \class         LagLeadOperatorTransformer
+///  \brief         Copy values from prior data or future data
+///                 Input type is a tuple of std::vector<std::string>, representing grains, and std::double_t, representing target column
+///
+template <typename InputTupleT>
```
michaelgsharp: This transformer isn't grain aware right? Isn't the grain wrapper going to do that for us? So isn't this type not a tuple and just the target value?

src/Featurizers/Components/WindowFeaturizerBase.h
angryjinyan: Yep. I just added it so it could compile for now. I guess this method and change in GrainEstimator constructor can be both included in a separate PR.

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,237 @@
...
+        }
+    }
+};
+
+/////////////////////////////////////////////////////////////////////////
+///  \typedef       LagLeadOperatorEstimator
+///  \brief         Estimator that creates `LagLeadOperatorTransformer`.
+///
+template <typename InputTupleT>
+using LagLeadOperatorEstimator           = Components::InferenceOnlyEstimatorImpl<LagLeadOperatorTransformer<InputTupleT>>;
```
michaelgsharp: You are going to have the same issue as me with this. The InferenceOnlyEstimatorImpl can't pass constructor parameters to the Transformer. I am currently working on fixing the grain transformer. Do you think I should also see if I can apply that fix to this one as well? If not, you will have to do what I did and actually make a full estimator that just does nothing.

src/Featurizers/LagLeadOperatorFeaturizer.h
angryjinyan: Oh I think you are right. The input type should just be the target value type.

src/Featurizers/LagLeadOperatorFeaturizer.h
angryjinyan: Per offline discussion, GrainEstimator change will be in a separate PR.

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,246 @@
...
+/////////////////////////////////////////////////////////////////////////
+///  \class         LagLeadOperatorTransformer
+///  \brief         Copy values from prior data or future data
+///                 Input type is a tuple of std::vector<std::string>, representing grains, and std::double_t, representing target column
+///
+template <typename InputT>
+class LagLeadOperatorTransformer :
+    public Transformer<
+        InputT,
+        Microsoft::Featurizer::RowMajMatrix<std::double_t>
```
davidbrownellWork: ><std::double_t [](start = 43, length = 14)

This should be InputT.

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,246 @@
...
+    void save(Archive &ar) const override;
+
+private:
+    // ----------------------------------------------------------------------
+    // |
+    // |  Private Types
+    // |
+    // ----------------------------------------------------------------------
+    using GrainType  = std::vector<std::string>;
+    // TODO: change target type to general types
```
davidbrownellWork: >TODO [](start = 7, length = 4)

Do this.

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,246 @@
...
+
+    // ----------------------------------------------------------------------
+    // |
+    // |  Private Data
+    // |
+    // ----------------------------------------------------------------------
+    std::uint32_t const                     _horizon;
+    std::vector<std::int64_t> const         _offsets;
+    Components::CircularBuffer<TargetType>  _buffer;
+    std::int64_t const                      _max_future_offset;
```
davidbrownellWork: >std::int64_t const                      _max_future_offset; [](start = 4, length = 59)

Nit: declare invariants (const members) before stateful members.

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,246 @@
...
+
+    // ----------------------------------------------------------------------
+    // |
+    // |  Private Methods
+    // |
+    // ----------------------------------------------------------------------
+
+    // MSVC runs into problems when the declaration and definition are separated
+    void execute_impl(typename BaseType::InputType const &input, typename BaseType::CallbackFunction const &callback) override {
+        if (_max_future_offset != 0 && _index <= _max_future_offset + _horizon - 1) {
```
davidbrownellWork: >if (_max_future_offset != 0 && _index <= _max_future_offset + _horizon - 1) { [](start = 8, length = 77)

This clause needs comments. I believe that this code is caching the value because we need more data before producing the matrix?

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,246 @@
...
+            ++_numPending;
+            _buffer.push(input);
+            return;
+        }
+        Microsoft::Featurizer::RowMajMatrix<std::double_t> ret(_offsets.size(), _horizon);
+        for (std::uint32_t row = 0; row < _offsets.size(); ++row) {
+            // col ranges from _horizon - 1 to 0
+            for (std::uint32_t col = _horizon - 1; col < _horizon; --col) {
+                if ((_index - _numPending - col + _offsets[row] < 0)) {
+                    ret(static_cast<std::int32_t>(row), static_cast<std::int32_t>(_horizon - 1 - col)) = std::nan("");
```
davidbrownellWork: >std::nan("") [](start = 105, length = 12)

Use `Traits<InputT>::CreateNullableType()`.

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,246 @@
...
+    
+
+    // ----------------------------------------------------------------------
+    // |
+    // |  Private Methods
+    // |
+    // ----------------------------------------------------------------------
+
+    // MSVC runs into problems when the declaration and definition are separated
+    void execute_impl(typename BaseType::InputType const &input, typename BaseType::CallbackFunction const &callback) override {
```
davidbrownellWork: >execute_impl [](start = 9, length = 12)

I think that this code is too complicated for what it is doing. We can talk in person on how it can be simplified.

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,246 @@
...
+                if (horizon == 0) {
+                    throw std::invalid_argument("Horizon cannot be 0!");
+                }
+                return horizon;
+            }()
+    )),
+    _offsets(
+        std::move(
+            [&offsets](void) -> std::vector<std::int64_t> & {
+                if (offsets.size() == 0) {
```
davidbrownellWork: >offsets.size() == 0 [](start = 20, length = 19)

`offsets.empty()`

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,246 @@
...
+                    throw std::invalid_argument("Offsets is empty!");
+                }
+                return offsets;
+            }()
+        )
+    ),
+    // if there are more than one offset provided, we need find the difference between max and min value to determine buffer size
+    // and if there's only one offset, we use the absolute value of it as the size
+    _buffer(Components::CircularBuffer<TargetType>(
```
davidbrownellWork: >Components::CircularBuffer<TargetType> [](start = 12, length = 38)

As written, this code creates a temporary `CircularBuffer` type and then calls `_buffer`'s move constructor. Remove the class name to call the standard constructor.

src/Featurizers/UnitTests/LagLeadOperatorFeaturizer_UnitTests.cpp
```diff
@@ -0,0 +1,270 @@
...
+#include "../TestHelpers.h"
+
+namespace NS = Microsoft::Featurizer;
+
+#if (defined __clang__)
+#   pragma clang diagnostic push
+#   pragma clang diagnostic ignored "-Wfloat-equal"
+#endif
+
+TEST_CASE("Invalid constructor parameter") {
```
davidbrownellWork: >TEST_CASE [](start = 0, length = 9)

I would expect to see many more tests given the complexity of the code. There should be tests that exercise each branch in the logic.

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,246 @@
...
+namespace Featurizer {
+namespace Featurizers {
+
+/////////////////////////////////////////////////////////////////////////
+///  \class         LagLeadOperatorTransformer
+///  \brief         Copy values from prior data or future data
+///                 Input type is a tuple of std::vector<std::string>, representing grains, and std::double_t, representing target column
+///
+template <typename InputT>
+class LagLeadOperatorTransformer :
```
wangyems: Please add these for transformer as ort need them to allocate memory for matrix:
std::ptrdiff_t getRowsNumber() const {return _offsets.size();}
std::ptrdiff_t getColsNumber() const {return _horizon;}

src/Featurizers/UnitTests/LagLeadOperatorFeaturizer_UnitTests.cpp
```diff
@@ -0,0 +1,270 @@
...
+    using InputType = std::double_t;using InputType = std::double_t;
+    CHECK_THROWS_WITH(NS::Featurizers::LagLeadOperatorTransformer<InputType>(0, {-1}), "Horizon cannot be 0!");
+    std::vector<std::int64_t> offset{};
+    CHECK_THROWS_WITH(NS::Featurizers::LagLeadOperatorTransformer<InputType>(1, offset), "Offsets is empty!");
+}
+
+TEST_CASE("Test with one grain - horizon 2, lag 0") {
+    using InputType = std::double_t;
+    using TransformedType = NS::ColMajMatrix<std::double_t>;
```
wangyems: Row?

src/Featurizers/UnitTests/LagLeadOperatorFeaturizer_UnitTests.cpp
```diff
@@ -0,0 +1,270 @@
...
+    CHECK(ret[6](0,0) == 14);
+    CHECK(ret[6](0,1) == 15);
+    transformer.execute(17, callback);
+    CHECK(ret[7](0,0) == 15);
+    CHECK(ret[7](0,1) == 16);
+}
+
+TEST_CASE("Test with one grain - horizon 1, lag 2 lag 1") {
+    using InputType = std::double_t;
+    using TransformedType = NS::ColMajMatrix<std::double_t>;
```
wangyems: same

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,224 @@
...
+/////////////////////////////////////////////////////////////////////////
+///  \class         LagLeadOperatorTransformer
+///  \brief         Copy values from prior data or future data
+///                 Input type is a tuple of std::vector<std::string>, representing grains, and InputT, representing target column
+///
+template <typename InputT>
+class LagLeadOperatorTransformer :
+    public Transformer<
+        InputT,
+        Microsoft::Featurizer::RowMajMatrix<nonstd::optional<InputT>>
```
davidbrownellWork: >nonstd::optional<InputT> [](start = 44, length = 24)

use `typename Traits<InputT>::nullable_type`, as an `InputT` of `double` should not be wrapped in `nonstd::optional`.

src/Featurizers/LagLeadOperatorFeaturizer.h
davidbrownellWork: This solution won't scale to dynamically shaped matrix types; we need to design this in a more scalable way.

---
In reply to: [395172177](https://github.com/microsoft/FeaturizersLibrary/pull/171#discussion_r395172177) [](ancestors = 395172177)

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,224 @@
...
+
+    void save(Archive &ar) const override;
+
+private:
+    // ----------------------------------------------------------------------
+    // |
+    // |  Private Types
+    // |
+    // ----------------------------------------------------------------------
+    using GrainType  = std::vector<std::string>;
```
davidbrownellWork: >using GrainType  = std::vector<std::string>; [](start = 4, length = 44)

This object shouldn't be grain-aware.

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,224 @@
...
+namespace Featurizer {
+namespace Featurizers {
+
+/////////////////////////////////////////////////////////////////////////
+///  \class         LagLeadOperatorTransformer
+///  \brief         Copy values from prior data or future data
+///                 Input type is a tuple of std::vector<std::string>, representing grains, and InputT, representing target column
+///
+template <typename InputT>
+class LagLeadOperatorTransformer :
```
davidbrownellWork: >LagLeadOperatorTransformer [](start = 6, length = 26)

This should follow naming conventions similar to RollingWindowTransformer/Estimator when it comes to grains.

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,224 @@
...
+    using TargetType = nonstd::optional<InputT>;
+
+    // ----------------------------------------------------------------------
+    // |
+    // |  Private Data
+    // |
+    // ----------------------------------------------------------------------
+    std::uint32_t const                     _horizon;
+    std::vector<std::int64_t> const         _offsets;
+    std::int64_t const                      _max_future_offset;
```
davidbrownellWork: >std::int64_t const                      _max_future_offset; [](start = 4, length = 59)

This is essentially a cached value that is used in 2 places: when determining the buffer size during construction and during flush. Does it make sense to cache this or calculate it in those 2 places?

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,224 @@
...
+    // |  Private Methods
+    // |
+    // ----------------------------------------------------------------------
+    // MSVC runs into problems when the declaration and definition are separated
+    void execute_impl(typename BaseType::InputType const &input, typename BaseType::CallbackFunction const &callback) override {
+        execute_helper(nonstd::optional<typename BaseType::InputType>(input), callback);
+    }
+
+    void flush_impl(typename BaseType::CallbackFunction const & callback) override {
+        std::int64_t _numPending = 0;
```
davidbrownellWork: >_numPending [](start = 21, length = 11)

A prefix of `_` indicates that the variable is a class member. This is just a standard var, so it shouldn't have that prefix.

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,224 @@
...
+        while (_numPending < _max_future_offset) {
+            execute_helper(Traits<nonstd::optional<InputT>>::CreateNullValue(), callback);
```
davidbrownellWork: >execute_helper [](start = 12, length = 14)

Can this call `execute` directly? That would eliminate the need of `execute_helper`.

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,224 @@
...
+
+    void execute_helper(nonstd::optional<typename BaseType::InputType> const &input, typename BaseType::CallbackFunction const &callback);
+};
+
+/////////////////////////////////////////////////////////////////////////
+///  \typedef       LagLeadOperatorEstimator
+///  \brief         Estimator that creates `LagLeadOperatorTransformer`.
+///
+template <typename InputT>
+using LagLeadOperatorEstimator           = Components::InferenceOnlyEstimatorImpl<LagLeadOperatorTransformer<InputT>>;
```
davidbrownellWork: >LagLeadOperatorEstimator [](start = 6, length = 24)

This needs to be a class to pass the estimator name to the base class. See FromStringEstimator.h for an example.

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,224 @@
...
+    _buffer(
+        _horizon + 
+        (_offsets.size() == 1 ? 
+        static_cast<size_t>(std::abs(_offsets[0])) : 
+        static_cast<size_t>(_max_future_offset-_max_prior_offset))
+    ) {
+        // prepopulate circular buffer with null values to imitate non-existing prior rows
+        for (std::int64_t i = 0; i < (static_cast<std::int64_t>(_horizon) - _max_prior_offset - 1); ++i) {
+            _buffer.push(Traits<nonstd::optional<T>>::CreateNullValue());
+        }
```
davidbrownellWork: This code needs to be executed after a flush as well.

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,224 @@
...
+    _max_prior_offset(std::move(
+        *std::min_element(_offsets.cbegin(), _offsets.cend()) < 0 ?  *std::min_element(_offsets.cbegin(), _offsets.cend()) : 0
+    )),
+    // if there are more than one offset provided, we need find the difference between max and min value to determine buffer size
+    // and if there's only one offset, we use the absolute value of it as the size
+    _buffer(
+        _horizon + 
+        (_offsets.size() == 1 ? 
+        static_cast<size_t>(std::abs(_offsets[0])) : 
+        static_cast<size_t>(_max_future_offset-_max_prior_offset))
```
davidbrownellWork: This check doesn't work in all scenarios; for example, what happens if the caller incorrectly passes in `[-2, -2]` or `[0]`. This needs to be more resilient to handle all scenarios. Also, tests need to be added to verify these scenarios.

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,224 @@
...
+    Traits<std::vector<std::int64_t>>::serialize(ar, _offsets);
+
+    // Note that we aren't serializing working state
+}
+
+template <typename T>
+void LagLeadOperatorTransformer<T>::execute_helper(nonstd::optional<typename BaseType::InputType> const &input, typename BaseType::CallbackFunction const &callback) {
+    _buffer.push(input);
+    // until the circular buffer is full, we don't have enough data to generate output matrix
+    if (_buffer.is_full()) {
```
davidbrownellWork: >if (_buffer.is_full()) [](start = 4, length = 22)

Nit: return early to avoid unnecessary indentation.

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,224 @@
...
+            // so the difference between offsets[row] and _max_prior_offset is the position to find the data to copy
+            std::tuple<typename Components::CircularBuffer<TargetType>::iterator, typename Components::CircularBuffer<TargetType>::iterator> range = _buffer.range(_horizon, static_cast<size_t>(_offsets[row] - _max_prior_offset));
+            typename Components::CircularBuffer<TargetType>::iterator start_iter = std::get<0>(range);
+            for (std::int32_t col = 0; col < static_cast<std::int32_t>(_horizon); ++col) {
+                ret(static_cast<std::int32_t>(row), col) = *start_iter;
+                ++start_iter;
+            }
+        }
+        callback(ret);
+    }
```
davidbrownellWork: Nit: Add vertical whitespace to help provide context. At the very least, add white space between variable declarations and code.

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,224 @@
...
+
+    // Note that we aren't serializing working state
+}
+
+template <typename T>
+void LagLeadOperatorTransformer<T>::execute_helper(nonstd::optional<typename BaseType::InputType> const &input, typename BaseType::CallbackFunction const &callback) {
+    _buffer.push(input);
+    // until the circular buffer is full, we don't have enough data to generate output matrix
+    if (_buffer.is_full()) {
+        Microsoft::Featurizer::RowMajMatrix<nonstd::optional<T>> ret(_offsets.size(), _horizon);
```
davidbrownellWork: >nonstd::optional<T> [](start = 44, length = 19)

Use Traits.

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,224 @@
...
+        for (std::uint32_t row = 0; row < _offsets.size(); ++row) {
```
davidbrownellWork: >row [](start = 27, length = 3)

Rather than calling this `row`, call it `offset` as that is the language that you have already established with the class' constructor and member variables. That will eliminate the need for the comment on line 209.

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,224 @@
...
+            // the circular buffer starts with the needed data for the _max_prior_offset
+            // so the difference between offsets[row] and _max_prior_offset is the position to find the data to copy
+            std::tuple<typename Components::CircularBuffer<TargetType>::iterator, typename Components::CircularBuffer<TargetType>::iterator> range = _buffer.range(_horizon, static_cast<size_t>(_offsets[row] - _max_prior_offset));
```
davidbrownellWork: >_offsets[row] - _max_prior_offset [](start = 193, length = 33)

Rather than performing this calculation on every call, would it be possible to do it when assigning offsets in the constructor? That may eliminate the need to cache `_max_prior_offset` as well.

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,224 @@
...
+            typename Components::CircularBuffer<TargetType>::iterator start_iter = std::get<0>(range);
+            for (std::int32_t col = 0; col < static_cast<std::int32_t>(_horizon); ++col) {
+                ret(static_cast<std::int32_t>(row), col) = *start_iter;
+                ++start_iter;
+            }
+        }
+        callback(ret);
```
davidbrownellWork: >ret [](start = 17, length = 3)

`std::move(ret)` to avoid a copy.

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,224 @@
...
+    void save(Archive &ar) const override;
+
+private:
+    // ----------------------------------------------------------------------
+    // |
+    // |  Private Types
+    // |
+    // ----------------------------------------------------------------------
+    using GrainType  = std::vector<std::string>;
+    using TargetType = nonstd::optional<InputT>;
```
davidbrownellWork: >nonstd::optional [](start = 23, length = 16)

Use Traits.

src/Featurizers/LagLeadOperatorFeaturizer.h
davidbrownellWork: >Microsoft::Featurizer::RowMajMatrix<nonstd::optional<T>> [](start = 8, length = 56)

Use `typename BaseType::TransformedType` so we don't have to define this in multiple places.

src/Featurizers/LagLeadOperatorFeaturizer.h
davidbrownellWork: >range = _buffer.range(_horizon, static_cast<size_t>(_offsets[row] - _max_prior_offset)); [](start = 141, length = 88)

Most (if not all) modern compilers are able to figure this out, but it is similar to a mistake that the compiler isn't smart enough to always figure out....

Taken literally, this assignment will result in the following actions:

1) Default construct range
2) Create a temporary for the result of `_buffer.range`
3) Use range's move assignment operator to move the temporary

This is similar to initializing a member var like this:

`_myFoo(MyFoo(args))`

rather than

`myFoo(args)`.

When assigning vars, you should use one of these syntaxes:

A) `range(_buffer.range(_horizon, ...));` [older style used before initializer lists]
B) `range{_buffer.range(horizon, ...)}` [newer style]

src/Featurizers/LagLeadOperatorFeaturizer.h
angryjinyan: Good point. Changed to nullable_type.

src/Featurizers/LagLeadOperatorFeaturizer.h
angryjinyan: I don't think so because say the target has type std::string, execute_impl would have input type as std::string while flush_impl() need a method that takes in nonstd::optional<std::string> to pass a null value.

src/Featurizers/LagLeadOperatorFeaturizer.h
angryjinyan: Good to know! Changed range initialization to newer style

angryjinyan: I think moving the calculation to the constructor is a good idea so instead of keeping the real lag lead offsets, I can just cache relative difference according to `_max_prior_offset`. But I still need to cache `_max_prior_offset` in this approach, right?

src/Featurizers/LagLeadOperatorFeaturizer.h
angryjinyan: I think I introduce the concept of `row` because I need `row` to populate the output matrix. If I call it `offset`, wouldn't it be confusing for somebody to read `output_matrix[offset, col] = ...` ?

src/Featurizers/LagLeadOperatorFeaturizer.h
angryjinyan: This estimator will be changed to a real class that would take input parameters in constructor.

src/Featurizers/LagLeadOperatorFeaturizer.h
wangyems: sure

src/Featurizers/LagLeadOperatorFeaturizer.h
angryjinyan: Good point! Executed after flush() now.

src/Featurizers/LagLeadOperatorFeaturizer.h
angryjinyan: Changed the implementation and added unit tests for these scenarios.

src/Tools/CodeGenerator/Featurizers.yaml
```diff
@@ -1557,10 +1557,10 @@ featurizers:
       The resulting matrix size is K rows x N cols, where K is the number of offsets and N is the horizon.
 
       Horizon and offsets should be passed in during construction. Offsets are passed in as a vector of ints so multiple lag orders can be applied within one featurizer call.
-      Output type is a matrix of double where rows are grouped by different offsets and columns are grouped by horizon.
+      Output type is a matrix of optional<T> where rows are grouped by different offsets and columns are grouped by horizon.
 
       C++-style pseudo signature:
-        template <typename T> matrix<double> execute(std::vector<std::string> const &, T const &> const &value);
+        template <typename T> matrix<T?> execute(std::vector<std::string> const &, T const &> const &value);
```
wangyems: tuple?

angryjinyan: The return type should just be a matrix<nonstd::optional<T>>

wangyems: I mean the input type:
std::vector<std::string> const &, T const &>
looks missing tuple?

angryjinyan: Oh! That is correct. Added tuple.

src/Featurizers/Components/WindowFeaturizerBase.h
```diff
@@ -344,6 +346,12 @@ typename CircularBuffer<T>::const_iterator CircularBuffer<T>::cend() const {
     return end_impl<const_iterator>();
 }
 
+template <class T>
+void CircularBuffer<T>::clear () {
+    _start_offset = 0;
+    _data.clear();
```
davidbrownellWork: >_data.clear(); [](start = 4, length = 14)

When data is cleared from a vector, I believe that its capacity is reset as well. We probably want to reserve space as what is currently being done in the constructor. In fact, the constructor should probably call this method directly and remove any duplicated code.

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,412 @@
...
+    void execute_impl(typename BaseType::InputType const &input, typename BaseType::CallbackFunction const &callback) override {
+        execute_helper(TransformedTargetType(input), callback);
+    }
+
+    void flush_impl(typename BaseType::CallbackFunction const & callback) override {
+        std::int64_t numPending = 0;
+        std::int64_t max_future((_offset_delta + *std::max_element(_offsets.cbegin(), _offsets.cend())) > 0 ? (_offset_delta + *std::max_element(_offsets.cbegin(), _offsets.cend())) : 0);
+
+        while (numPending < max_future) {
+            execute_helper(Traits<TransformedTargetType>::CreateNullValue(), callback);
```
davidbrownellWork: This could could use come comments to indicate that we are adding null values to ensure that we generate output for all input previously received.

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,412 @@
...
+            ++numPending;
+        }
+        // after flush is called, this transfomer should be reset to its starting state
+        _buffer.clear();
+        for (std::int64_t i = 0; i < (static_cast<std::int64_t>(_horizon) - _offset_delta - 1); ++i) {
+            _buffer.push(Traits<TransformedTargetType>::CreateNullValue());
+        }
```
davidbrownellWork: This code is duplicated in the ctor as well. Implement it in a single place.

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,412 @@
...
+                    throw std::invalid_argument("Offsets is empty!");
+                }
+                return *std::min_element(offsets.cbegin(), offsets.cend()) < 0 ? *std::min_element(offsets.cbegin(), offsets.cend()) : 0;
+            }()
+        )
+    ),
+    // calibrate offsets to minimum offset = 0
+    _offsets(
+        std::move(
+            [&offsets](void) -> std::vector<std::int64_t> & {
```
davidbrownellWork: >offsets [](start = 14, length = 7)

Error if offsets is empty.

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,412 @@
...
+                return horizon;
+            }()
+    )),
+    _offset_delta(
+        std::move(
+            [&offsets](void) -> std::int64_t {
+                if (offsets.size() == 0) {
+                    throw std::invalid_argument("Offsets is empty!");
+                }
+                return *std::min_element(offsets.cbegin(), offsets.cend()) < 0 ? *std::min_element(offsets.cbegin(), offsets.cend()) : 0;
```
davidbrownellWork: >std::min_element(offsets.cbegin(), offsets.cend()) < 0 ? *std::min_element(offsets.cbegin(), offsets.cend()) : 0 [](start = 24, length = 112)

Implement this without enumerating the list of offsets 2 times.

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,412 @@
...
+            }()
+        )
+    ),
+    // calibrate offsets to minimum offset = 0
+    _offsets(
+        std::move(
+            [&offsets](void) -> std::vector<std::int64_t> & {
+                std::int64_t min = *std::min_element(offsets.cbegin(), offsets.cend()) < 0 ? *std::min_element(offsets.cbegin(), offsets.cend()) : 0;
```
davidbrownellWork: >std::min_element(offsets.cbegin(), offsets.cend()) < 0 ? *std::min_element(offsets.cbegin(), offsets.cend()) : 0 [](start = 36, length = 112)

Implement this without enumerating the list of offsets 2 times.

src/Featurizers/LagLeadOperatorFeaturizer.h
angryjinyan: If offsets is empty, it will be caught in line 226.

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,412 @@
...
+    // if it is, the buffer size is the absolute value of minimum offset plus horizon
+    // if it's not, the buffer size is the difference between min and max of offsets, which is already calculated in the offsets calibration so we can use max_element directly
+    _buffer(
+        _horizon + 
+        (_offsets.size() == 1 ? 
+        static_cast<size_t>(std::abs(_offsets[0] + _offset_delta)) : 
+        static_cast<size_t>(
+            *std::max_element(_offsets.cbegin(), _offsets.cend()) + _offset_delta < 0 ? 
+            std::abs(_offset_delta) : 
+            *std::max_element(_offsets.cbegin(), _offsets.cend())))
```
davidbrownellWork: Implement this without enumerating the lists of offsets 2 times.

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,412 @@
...
+void LagLeadOperatorTransformer<T>::execute_helper(TransformedTargetType const &input, typename BaseType::CallbackFunction const &callback) {
+    _buffer.push(input);
+    // until the circular buffer is full, we don't have enough data to generate output matrix
+    if (!_buffer.is_full()) {
+        return;
+    }
+    typename BaseType::TransformedType ret(_offsets.size(), _horizon);
+    
+    for (std::uint32_t row = 0; row < _offsets.size(); ++row) {
+        std::tuple<typename Components::CircularBuffer<TransformedTargetType>::iterator, typename Components::CircularBuffer<TransformedTargetType>::iterator> range{_buffer.range(_horizon, static_cast<size_t>(_offsets[row]))};
```
davidbrownellWork: >static_cast<size_t>(_offsets[row]) [](start = 189, length = 34)

This cast is working around a potential issue. Conceptually speaking, an offset should never be > 0. A delta can be less than 0. So, incoming deltas are converted to offsets in the constructor. Deltas should be a signed type, offsets should not unsigned.

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,412 @@
...
+void LagLeadOperatorTransformer<T>::save(Archive &ar) const /*override*/ {
+    // Version
+    Traits<std::uint16_t>::serialize(ar, 1); // Major
+    Traits<std::uint16_t>::serialize(ar, 0); // Minor
+
+    // Data
+    Traits<std::uint32_t>::serialize(ar, _horizon);
+
+    // reset _offsets to its input version
+    std::vector<std::int64_t> offsets(_offsets);
```
davidbrownellWork: >offsets [](start = 30, length = 7)

deltas

src/Featurizers/UnitTests/LagLeadOperatorFeaturizer_UnitTests.cpp
```diff
@@ -0,0 +1,651 @@
...
+        }
+    );
+
+    transformer.execute(10, callback);
+    transformer.execute(11, callback);
+    transformer.execute(12, callback);
+    transformer.execute(13, callback);
+    transformer.execute(14, callback);
+    transformer.execute(15, callback);
+    transformer.flush(callback);
```
davidbrownellWork: >transformer.flush(callback); [](start = 4, length = 28)

Before the flush, we should have N items. Verify that is happening.

src/Featurizers/UnitTests/LagLeadOperatorFeaturizer_UnitTests.cpp
```diff
@@ -0,0 +1,651 @@
+// ----------------------------------------------------------------------
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License
+// ----------------------------------------------------------------------
+#define CATCH_CONFIG_MAIN
```
davidbrownellWork: >CATCH_CONFIG_MAIN [](start = 8, length = 17)

I believe that all of these tests invoke the functionality with deltas provided in sorted order; can you add a test where the incoming deltas are not in sorted order to verify things still work as expected.

src/Featurizers/Components/WindowFeaturizerBase.h
angryjinyan: clear() will be added in a separate PR.

src/Featurizers/LagLeadOperatorFeaturizer.h
angryjinyan: Comment added

src/Featurizers/LagLeadOperatorFeaturizer.h
angryjinyan: Created a separate function to prepopulate nulls.

src/Featurizers/LagLeadOperatorFeaturizer.h
angryjinyan: Good point! The transformer is now keeping offsets as uint32_t now.

src/Featurizers/LagLeadOperatorFeaturizer.h
angryjinyan: I'm not sure I understand this comment.

davidbrownellWork: The concept of deltas (signed values) and offsets (unsigned values) are being mixed up. In this case, you are serializing deltas rather than offsets.

angryjinyan: Resolved per offline discussion. Changed names according to offset and delta concepts

src/Featurizers/UnitTests/LagLeadOperatorFeaturizer_UnitTests.cpp
angryjinyan: Added a test that deltas are not in sorted order

src/Featurizers/LagLeadOperatorFeaturizer.h
```diff
@@ -0,0 +1,423 @@
...
+    // |
+    // ----------------------------------------------------------------------
+    // MSVC runs into problems when the declaration and definition are separated
+    void execute_impl(typename BaseType::InputType const &input, typename BaseType::CallbackFunction const &callback) override {
+        execute_helper(TransformedTargetType(input), callback);
+    }
+
+    void flush_impl(typename BaseType::CallbackFunction const & callback) override {
+        std::int64_t numPending(0);
+        std::int64_t max_future(_offset_delta + std::max(static_cast<std::int64_t>(*std::max_element(_offsets.cbegin(), _offsets.cend())), static_cast<std::int64_t>(0)));
```
davidbrownellWork: >max_future [](start = 21, length = 10)

I don't think that this value will ever be < 0 will it? If not, make it unsigned. If this is unsigned, numPending can be unsigned as well.

angryjinyan: It can be a negative value. Basically, this value is used for do I need to execute more NULLs and if the input deltas are, say {-3, -1}, max_future would be -1.

PR #172, opened 2020-03-19
CircularBuffer add in size() method

angryjinyan:


src/Featurizers/Components/WindowFeaturizerBase.h
```diff
@@ -135,6 +135,8 @@ namespace Components {
         // ----------------------------------------------------------------------
         CircularBuffer(size_t max_size);
 
+        size_t size(void) const;
```
yuslepukhin: >void [](start = 20, length = 4)

void seems out of place, everything else in this file does not have it.
In C++ (versus C) empty arguments list is the same as void.

angryjinyan: Good point! void removed.

PR #173, opened 2020-03-19
Analytical Rolling Window Featurizer

michaelgsharp:
Transformer is working. Still not supporting by grain.


src/Featurizers/AnalyticalRollingWindowFeaturizer.h
```diff
@@ -0,0 +1,307 @@
...
+    bool operator!=(AnalyticalRollingWindowTransformer const &other) const;
+    void save(Archive & ar) const override;
+
+private:
+    // ----------------------------------------------------------------------
+    // |
+    // |  Private Types
+    // |
+    // ----------------------------------------------------------------------
+    using OutputType = std::vector<double>;
```
davidbrownellWork: >OutputType [](start = 10, length = 10)

This is the same as `BaseType::TransformedType`.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
```diff
@@ -0,0 +1,307 @@
...
+    // |
+    // |  Private Members
+    // |
+    // ----------------------------------------------------------------------
+    // _windowSizes and _minWindowCount are int64 so that we can have the full positive range of the uint32 it comes
+    // in as but can still support negative numbers.
+    Components::CircularBuffer<InputT>      _buffer;
+    std::int64_t                            _windowSize;
+    std::uint32_t                           _horizon;
+    std::int64_t                            _minWindowCount;
```
davidbrownellWork: Some of these can probably be const.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
```diff
@@ -0,0 +1,307 @@
...
+    // |  Public Methods
+    // |
+    // ----------------------------------------------------------------------
+
+    // MSVC runs into problems when the declaration and definition are separated
+    void execute_impl(typename BaseType::InputType const &input, typename BaseType::CallbackFunction const &callback) override {
+        OutputType         results(_horizon);
+        
+        _buffer.push(input);
+        size_t bufferSize = _buffer.size();
```
davidbrownellWork: Nit: I'd move these statements to the top of the function; that way it clearly delineates between the "buffer management stuff" and the "result creation stuff".

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
davidbrownellWork: >std::int64_t [](start = 4, length = 12)

This should probably be an unsigned type since it doesn't make sense to have a negative minimum window count.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
```diff
@@ -0,0 +1,307 @@
...
+            // We cast to int32 here because -1 is valid for the algorithm.
+            std::int32_t elementsAvailableToUse(static_cast<std::int32_t>(bufferSize) - i);
+
+            // If we don't have enough elements then output NaN
+            if (elementsAvailableToUse < _minWindowCount) {
+                results[_horizon - i] = std::nan("");
+
+            // If we have strictly less elements then the window size, but more then the minimum size
+            } else if (elementsAvailableToUse < _windowSize) {
+                const auto range = _buffer.range(_windowSize - elementsAvailableToUse);
```
davidbrownellWork: >auto [](start = 22, length = 4)

Only use `auto` in these situations:

1) Variable defined within a loop (e.g. a for loop)
2) When declaring an anonymous function
3) In complicated template metaprogramming scenarios.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
```diff
@@ -0,0 +1,307 @@
...
+                results[_horizon - i] = Calculators::MeanCalculator<BaseType::InputType>::execute(start_iter, end_iter);
+
+            // More elements are in the buffer than are needed in the window
+            } else {
+                const auto range = _buffer.range(_windowSize, bufferSize - i - _windowSize);
+                auto start_iter = std::get<0>(range);
+                auto end_iter = std::get<1>(range);
+
+                results[_horizon - i] = Calculators::MeanCalculator<BaseType::InputType>::execute(start_iter, end_iter);
+            }
```
davidbrownellWork: It looks like the only difference between these 2 else blocks is how the range is calculated. Merge these 2 into a single else clause and use a ternary statement or a lambda function to properly calculate the range.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
```diff
@@ -0,0 +1,307 @@
...
+    // MSVC runs into problems when the declaration and definition are separated
+    void execute_impl(typename BaseType::InputType const &input, typename BaseType::CallbackFunction const &callback) override {
+        OutputType         results(_horizon);
+        
+        _buffer.push(input);
+        size_t bufferSize = _buffer.size();
+
+        for (int i = _horizon; i > 0; i--){
+            // We cast to int32 here because -1 is valid for the algorithm.
+            std::int32_t elementsAvailableToUse(static_cast<std::int32_t>(bufferSize) - i);
```
davidbrownellWork: >std::int32_t [](start = 12, length = 12)

Make this const to communicate that it won't be updated again.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
```diff
@@ -0,0 +1,307 @@
...
+
+            // If we don't have enough elements then output NaN
+            if (elementsAvailableToUse < _minWindowCount) {
+                results[_horizon - i] = std::nan("");
```
davidbrownellWork: >std::nan("") [](start = 40, length = 12)

`Traits<InputT>::CreateNullValue()`

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
```diff
@@ -0,0 +1,307 @@
...
+private:
+    // ----------------------------------------------------------------------
+    // |
+    // |  Private Members
+    // |
+    // ----------------------------------------------------------------------
+    std::uint32_t                           _windowSize;
+    std::uint32_t                           _horizon;
+    std::uint32_t                           _minWindowCount;
+    AnalyticalRollingWindowCalculation      _windowCalculation;
```
davidbrownellWork: These should all be const.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
```diff
@@ -0,0 +1,307 @@
...
+// |
+// ----------------------------------------------------------------------
+template <typename InputT, size_t MaxNumTrainingItemsV>
+AnalyticalRollingWindowTransformer<InputT, MaxNumTrainingItemsV>::AnalyticalRollingWindowTransformer(std::uint32_t windowSize, AnalyticalRollingWindowCalculation windowCalculation, std::uint32_t horizon, std::uint32_t minWindowCount) :
+    _windowSize(std::move(windowSize)),
+    _windowCalculation(std::move(windowCalculation)),
+    _horizon(std::move(horizon)),
+    _minWindowCount(std::move(minWindowCount)),
+    _buffer(horizon + (windowSize)) {
+
```
davidbrownellWork: Validate these values before using them.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
```diff
@@ -0,0 +1,307 @@
...
+// |  AnalyticalRollingWindowEstimator
+// |
+// ----------------------------------------------------------------------
+template <typename InputT, size_t MaxNumTrainingItemsV>
+AnalyticalRollingWindowEstimator<InputT, MaxNumTrainingItemsV>::AnalyticalRollingWindowEstimator(AnnotationMapsPtr pAllColumnAnnotations, std::uint32_t windowSize, AnalyticalRollingWindowCalculation windowCalculation, std::uint32_t horizon, std::uint32_t minWindowCount) :
+    BaseType(AnalyticalRollingWindowEstimatorName, std::move(pAllColumnAnnotations)),
+    _windowSize(std::move(windowSize)),
+    _windowCalculation(std::move(windowCalculation)),
+    _horizon(std::move(horizon)),
+    _minWindowCount(std::move(minWindowCount))
```
davidbrownellWork: Validate these values before using them.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
michaelgsharp: Good catch. I have removed it and am using BaseType instead.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
michaelgsharp: All of them but the buffer itself can. I have added const to them.

michaelgsharp: I did that so I would have to cast it less, but after our earlier discussion the casting isn't necesary anymore so they are all back to `uint32`

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
michaelgsharp: So this auto is in a for loop. I have now replaced it with the actual type anyways. Interesting though is that in the process I ran into an internal compiler error. I was able to work around it, it was just interesting to see.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
michaelgsharp: Resolving this based on our offline discussion.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
michaelgsharp: This has been removed entirely.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
michaelgsharp: I can't use `InputT` here as the input type doesn't matter, its always going to double. I have changed it to be `Traits<double>::CreateNullValue()`. I could do a mapping here if needed, though in this case it would always be InputT to double. If you have any other thoughts on that let me know.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
michaelgsharp: All have been changed to const.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
michaelgsharp: With something like this is it best to validate in the Estimator, the Transformer, or both? I think just the Estimator would work. Anything I am missing with that assumption?

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
```diff
@@ -0,0 +1,305 @@
...
+    // |
+    // ----------------------------------------------------------------------
+    using BaseType = Components::InferenceOnlyTransformerImpl<InputT, std::vector<double>>;
+
+    // ----------------------------------------------------------------------
+    // |
+    // |  Public Methods
+    // |
+    // ----------------------------------------------------------------------
+    AnalyticalRollingWindowTransformer(std::uint32_t windowSize, AnalyticalRollingWindowCalculation windowCalculation, std::uint32_t horizon, std::uint32_t minWindowCount = 1) ;
```
davidbrownellWork: >minWindowCount [](start = 156, length = 14)

minWindowSize

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
```diff
@@ -0,0 +1,305 @@
...
+    // |
+    // |  Private Members
+    // |
+    // ----------------------------------------------------------------------
+    // _windowSizes and _minWindowCount are int64 so that we can have the full positive range of the uint32 it comes
+    // in as but can still support negative numbers.
+    Components::CircularBuffer<InputT>              _buffer;
+    const std::uint32_t                             _windowSize;
+    const std::uint32_t                             _horizon;
+    const std::uint32_t                             _minWindowSize;
```
davidbrownellWork: >const std::uint32_t                             _minWindowSize; [](start = 4, length = 63)

Nit: I'd move this closer to the declaration of `_windowSize` since they are related.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
```diff
@@ -0,0 +1,305 @@
...
+private:
+    // ----------------------------------------------------------------------
+    // |
+    // |  Private Members
+    // |
+    // ----------------------------------------------------------------------
+    // _windowSizes and _minWindowCount are int64 so that we can have the full positive range of the uint32 it comes
+    // in as but can still support negative numbers.
+    Components::CircularBuffer<InputT>              _buffer;
+    const std::uint32_t                             _windowSize;
```
davidbrownellWork: >_windowSize [](start = 52, length = 11)

Nit: `_maxWindowSize`?

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
```diff
@@ -0,0 +1,305 @@
...
+
+private:
+    // ----------------------------------------------------------------------
+    // |
+    // |  Private Members
+    // |
+    // ----------------------------------------------------------------------
+    // _windowSizes and _minWindowCount are int64 so that we can have the full positive range of the uint32 it comes
+    // in as but can still support negative numbers.
+    Components::CircularBuffer<InputT>              _buffer;
```
davidbrownellWork: >Components::CircularBuffer<InputT>              _buffer; [](start = 4, length = 56)

Nit: Declare class invariants before mutable data.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
```diff
@@ -0,0 +1,305 @@
...
+    // |  Public Methods
+    // |
+    // ----------------------------------------------------------------------
+
+    // MSVC runs into problems when the declaration and definition are separated
+    void execute_impl(typename BaseType::InputType const &input, typename BaseType::CallbackFunction const &callback) override {
+        typename BaseType::TransformedType         results(_horizon);
+        
+        _buffer.push(input);
+        size_t bufferSize = _buffer.size();
```
davidbrownellWork: >size_t [](start = 8, length = 6)

Make this const since it is never updated again.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
```diff
@@ -0,0 +1,305 @@
...
+
+    // ----------------------------------------------------------------------
+    // |
+    // |  Public Methods
+    // |
+    // ----------------------------------------------------------------------
+
+    // MSVC runs into problems when the declaration and definition are separated
+    void execute_impl(typename BaseType::InputType const &input, typename BaseType::CallbackFunction const &callback) override {
+        typename BaseType::TransformedType         results(_horizon);
```
davidbrownellWork: >typename BaseType::TransformedType         results(_horizon); [](start = 8, length = 61)

Declare this closer to the code where it is used.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
```diff
@@ -0,0 +1,305 @@
...
+        
+        _buffer.push(input);
+        size_t bufferSize = _buffer.size();
+
+        double result;
+        size_t numElements;
+        size_t startingOffset;
```
davidbrownellWork: These variables are only used within the scope of the for loop and should be defined there.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
```diff
@@ -0,0 +1,305 @@
...
+
+        for (std::uint32_t offset = 0; offset < _horizon; ++offset){
+
+            // If we don't have enough elements then output NaN
+            if ((_horizon - offset) + _minWindowSize > bufferSize) {
```
davidbrownellWork: >(_horizon - offset) + _minWindowSize > bufferSize [](start = 16, length = 49)

Nit: When I read this, I have to keep an expression with 3 vars in my head to see how it compares to `bufferSize`. For me, it is more natural to read this as `if(bufferSize < _horizon - offset + _minWindowSize)`.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
```diff
@@ -0,0 +1,305 @@
...
+                result = Traits<double>::CreateNullValue();
+
+            // If we have strictly less elements then the window size, but more then the minimum size
```
davidbrownellWork: >// If we have strictly less elements then the window size, but more then the minimum size [](start = 12, length = 89)

This comment isn't accurate anymore.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
```diff
@@ -0,0 +1,305 @@
...
+            } else {
+                if (bufferSize - (_horizon - offset) < _windowSize) {
```
davidbrownellWork: >_horizon - offset [](start = 34, length = 17)

This concept is repeated in the code. I think that this is the desired windowSize? This would make more sense if `_windowSize` was renamed `_maxWindowSize`.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
```diff
@@ -0,0 +1,305 @@
...
+                    numElements = bufferSize - (_horizon - offset); //asfd
```
davidbrownellWork: >//asfd [](start = 68, length = 6)

Not sure what this means.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
```diff
@@ -0,0 +1,305 @@
...
+                    startingOffset = 0;
+                } else {
+                    numElements = _windowSize;
+                    startingOffset = offset - (_buffer.capacity() - bufferSize);
+                }
+
+                const std::tuple<Components::CircularBuffer<InputT>::iterator, Components::CircularBuffer<InputT>::iterator> range = _buffer.range(numElements, startingOffset);
```
davidbrownellWork: >const std::tuple<Components::CircularBuffer<InputT>::iterator, Components::CircularBuffer<InputT>::iterator> range = _buffer.range(numElements, startingOffset); [](start = 16, length = 160)

Most (if not all) modern compilers are able to figure this out, but it is similar to a mistake that the compiler isn't smart enough to always figure out....

Taken literally, this assignment will result in the following actions:

1) Default construct range
2) Create a temporary for the result of `_buffer.range`
3) Use range's move assignment operator to move the temporary

This is similar to initializing a member var like this:

`_myFoo(MyFoo(args))`

rather than

`myFoo(args)`

When assigning vars, you should use one of these syntaxes:

A) `range(_buffer.range(_horizon, ...));` [older style used before initializer lists]
B) `range{_buffer.range(_horizon, ...)};` [newer style]

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
```diff
@@ -0,0 +1,305 @@
...
+    // |
+    // ----------------------------------------------------------------------
+    using BaseType                          = TransformerEstimator<InputT, std::vector<double>>;
+    using TransformerType                   = AnalyticalRollingWindowTransformer<InputT, MaxNumTrainingItemsV>;
+    // ----------------------------------------------------------------------
+    // |
+    // |  Public Methods
+    // |
+    // ----------------------------------------------------------------------
+    AnalyticalRollingWindowEstimator(AnnotationMapsPtr pAllColumnAnnotations, std::uint32_t windowSize, AnalyticalRollingWindowCalculation windowCalculation, std::uint32_t horizon, std::uint32_t minWindowCount = 1);
```
davidbrownellWork: >minWindowCount [](start = 195, length = 14)

minWindowSize

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
```diff
@@ -0,0 +1,305 @@
...
+/////////////////////////////////////////////////////////////////////////
+///  \class         AnalyticalRollingWindowEstimator
+///  \brief         Estimator that creates `AnalyticalRollingWindowTransformer`.
+///
+template <
+    typename InputT,
+    size_t MaxNumTrainingItemsV=std::numeric_limits<size_t>::max()
+>
+class AnalyticalRollingWindowEstimator :
+    public TransformerEstimator<InputT, std::vector<double>> {
```
davidbrownellWork: >TransformerEstimator<InputT, std::vector<double>> [](start = 11, length = 49)

You can base this class on `Components::InferenceOnlyEstimatorImpl` will simplify the implementation.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
davidbrownellWork: Per our offline conversation: both.

---
In reply to: [395305722](https://github.com/microsoft/FeaturizersLibrary/pull/173#discussion_r395305722) [](ancestors = 395305722)

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
davidbrownellWork: By "Variable defined within a loop", I mean something like:

`for(auto it = container.begin(); it != container.end(); ++it) { ... }`

---
In reply to: [395302156](https://github.com/microsoft/FeaturizersLibrary/pull/173#discussion_r395302156) [](ancestors = 395302156)

michaelgsharp:
the Simple RollingWIndow isn't ready for review yet. I need to update it with the common base class still. Analytical is ready though. Simple will be updated shortly

src/Featurizers/Components/RollingWindowTransformerBase.h
```diff
@@ -0,0 +1,163 @@
+// ----------------------------------------------------------------------
```
davidbrownellWork: This should live in the Base subdir (which is a sibling of Components)

src/Featurizers/Components/RollingWindowTransformerBase.h
```diff
@@ -0,0 +1,163 @@
...
+    // MSVC runs into problems when the declaration and definition are separated
+    void execute_impl(typename BaseType::InputType const &input, typename BaseType::CallbackFunction const &callback) override {
+        typename BaseType::TransformedType         results(_horizon);
+        
+        _buffer.push(input);
+        size_t bufferSize = _buffer.size();
+
+        OutputT result;
+        size_t numElements;
+        size_t startingOffset;
```
davidbrownellWork: These aren't used outside of the loop so they should be defined local within the loop.

src/Featurizers/Components/RollingWindowTransformerBase.h
```diff
@@ -0,0 +1,163 @@
...
+    bool operator!=(RollingWindowTransformerBase const &other) const;
+
+protected:
+    // ----------------------------------------------------------------------
+    // |
+    // |  Protected Members
+    // |
+    // ----------------------------------------------------------------------
+    // These are protected so that the derived classes can use them for saving of state.
+    const std::uint32_t                             _windowSize;
```
davidbrownellWork: >_windowSize [](start = 52, length = 11)

I'd suggest renaming this to `_maxWindowSize`

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
michaelgsharp: Validation has been added.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
michaelgsharp: Validation has been added.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
michaelgsharp: Fixed. It was reverted by a git rebase, but its fixed again.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
michaelgsharp: I put it at the bottom since its optional in the constructor and thats is semi how I had them organized here as well. I will move it, but what is generally considered good practice for stuff like that?

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
michaelgsharp: Renamed.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
michaelgsharp: Reordered.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
michaelgsharp: added const modifier.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
michaelgsharp: Moved it down to right above the for loop.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
michaelgsharp: Moved them in the loop.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
michaelgsharp: It should have been moved down to the first `if` under that `else`. It has been moved.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
michaelgsharp: I have fixed this and the other if statement to be in the same format.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
michaelgsharp: I have renamed this everywhere to `_maxWindowSize`

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
michaelgsharp: It is now removed. It was a reminder for me to come back and fix it.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
michaelgsharp: Good to know. I have fixed this and the start/end iterators below it too.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
michaelgsharp: I had an issue when doing that. Let me try again and see.

src/Featurizers/Components/RollingWindowTransformerBase.h
michaelgsharp: Good point. Relocated the file.

src/Featurizers/Components/RollingWindowTransformerBase.h
michaelgsharp: Moved into the loop.

src/Featurizers/Components/RollingWindowTransformerBase.h
michaelgsharp: Renamed.

michaelgsharp:
It looks like a lot of your older comments got undone when I did a rebase (poorly) @davidbrownellWork . They should be resolved now.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
michaelgsharp: Resolved.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
```diff
@@ -0,0 +1,337 @@
...
+    // Note that we aren't serializing working state
+}
+
+// ----------------------------------------------------------------------
+// |
+// |  AnalyticalRollingWindowEstimator
+// |
+// ----------------------------------------------------------------------
+template <typename InputT, size_t MaxNumTrainingItemsV>
+AnalyticalRollingWindowEstimator<InputT, MaxNumTrainingItemsV>::AnalyticalRollingWindowEstimator(AnnotationMapsPtr pAllColumnAnnotations, std::uint32_t maxWindowSize, AnalyticalRollingWindowCalculation windowCalculation, std::uint32_t horizon, std::uint32_t minWindowSize) :
```
davidbrownellWork: With this change, it would probably make sense to have the `maxWindowSize` and `minWindowSize` parameters appear next to each other. Also the members should be next to each other as well.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
```diff
@@ -0,0 +1,337 @@
...
+                if(horizon < 1)
+                    throw std::invalid_argument("horizon");
+
+                return horizon;
+            }()
+        )
+    ),
+    _minWindowSize(
+        std::move(
+            [&minWindowSize]() -> std::uint32_t & {
```
davidbrownellWork: >minWindowSize [](start = 14, length = 13)

Ensure that `_minWindowSize` is < `_maxWindowSize`.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
```diff
@@ -0,0 +1,337 @@
...
+    // |
+    // ----------------------------------------------------------------------
+    using BaseType = Base::RollingWindowTransformerBase<InputT, double, MaxNumTrainingItemsV>;
+
+    // ----------------------------------------------------------------------
+    // |
+    // |  Public Methods
+    // |
+    // ----------------------------------------------------------------------
+    AnalyticalRollingWindowTransformer(std::uint32_t maxWindowSize, AnalyticalRollingWindowCalculation windowCalculation, std::uint32_t horizon, std::uint32_t minWindowSize = 1) ;
```
davidbrownellWork: >AnalyticalRollingWindowTransformer [](start = 4, length = 34)

Same comments here as in `AnalyticalRollingWindowEstimator`'s constructor.

src/Featurizers/Base/RollingWindowTransformerBase.h
```diff
@@ -0,0 +1,187 @@
...
+        _buffer.push(input);
+        const size_t bufferSize = _buffer.size();
+
+        typename BaseType::TransformedType         results(_horizon);
+
+        for (std::uint32_t offset = 0; offset < _horizon; ++offset){
+            OutputT result;
+            size_t numElements;
+            size_t startingOffset;
+
```
davidbrownellWork: These values are only used in the else clause - define them there. As a general rule, define variables as close as possible to where they are used. Bonus points if you can declare the vars as const.

src/Featurizers/Base/RollingWindowTransformerBase.h
```diff
@@ -0,0 +1,187 @@
...
+    Components::CircularBuffer<InputT>              _buffer;
+
+    // ----------------------------------------------------------------------
+    // |
+    // |  Public Methods
+    // |
+    // ----------------------------------------------------------------------
+
+    // MSVC runs into problems when the declaration and definition are separated
+    void execute_impl(typename BaseType::InputType const &input, typename BaseType::CallbackFunction const &callback) override {
```
davidbrownellWork: >execute_impl [](start = 9, length = 12)

Does this object need `flush_impl` to reset state?

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
```diff
@@ -0,0 +1,337 @@
...
+            Components::GrainEstimatorImpl<GrainType, AnalyticalRollingWindowEstimator<InputT, MaxNumTrainingItemsV>>,
+            Components::FilterDecoratorEstimator<std::tuple<GrainType const &, std::vector<double>>, 1>
+        >;
+    
+    // ----------------------------------------------------------------------
+    // |
+    // |  Public Methods
+    // |
+    // ----------------------------------------------------------------------
+    // MSVC has problems when the definition for the constructor is separated from its declaration.
```
davidbrownellWork: >// MSVC has problems when the definition for the constructor is separated from its declaration. [](start = 4, length = 95)

This comment can be removed.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
```diff
@@ -0,0 +1,337 @@
...
+            // Version
+            std::uint16_t                   majorVersion(Traits<std::uint16_t>::deserialize(ar));
+            std::uint16_t                   minorVersion(Traits<std::uint16_t>::deserialize(ar));
+
+            if(majorVersion != 1 || minorVersion != 0)
+                throw std::runtime_error("Unsupported archive version");
+
+            // Because we know that maxWindowSize and minWindowSize come in originally as uint32, this cast is always safe
+            std::uint32_t                       maxWindowSize(Traits<std::uint32_t>::deserialize(ar));
+            AnalyticalRollingWindowCalculation  windowCalculation(static_cast<AnalyticalRollingWindowCalculation>(Traits<std::uint8_t>::deserialize(ar)));
```
davidbrownellWork: >std::uint8_t [](start = 121, length = 12)

Nit: `typename std::underlying_type<AnalyticalRollingWindowCalculation>::type`

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
```diff
@@ -0,0 +1,337 @@
...
+    AnalyticalRollingWindowTransformer(
+        [&ar](void) {
+            // Version
+            std::uint16_t                   majorVersion(Traits<std::uint16_t>::deserialize(ar));
+            std::uint16_t                   minorVersion(Traits<std::uint16_t>::deserialize(ar));
+
+            if(majorVersion != 1 || minorVersion != 0)
+                throw std::runtime_error("Unsupported archive version");
+
+            // Because we know that maxWindowSize and minWindowSize come in originally as uint32, this cast is always safe
```
davidbrownellWork: >// Because we know that maxWindowSize and minWindowSize come in originally as uint32, this cast is always safe [](start = 12, length = 110)

Does this comment still apply?

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
michaelgsharp: The problem is that the `minWindowSize` parameter has a default value (in the declaration), so it needs to be at the end. Would it be better to move `maxWindowSize` next to it at the end then?

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
michaelgsharp: Check has been added.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
michaelgsharp: The checking for the transformer happens in the base class. Check has been added.

src/Featurizers/Base/RollingWindowTransformerBase.h
michaelgsharp: Yep. That is on my todo list, but I had forgotten about it this morning. I'll add it in now.

src/Featurizers/Base/RollingWindowTransformerBase.h
michaelgsharp: I moved them into the else clause, but since they have to be declared before they are initialized I can't make them const.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
michaelgsharp: Good catch. Removed.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
michaelgsharp: Switched the std::uint8_t for this instead.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
michaelgsharp: Nope, I will remove it.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
michaelgsharp: Moved them together. Resolving.

src/Tools/CodeGenerator/Featurizers.yaml
```diff
@@ -28,7 +28,7 @@ featurizers:
             | A         | 11    | [5]               |
             +-----------+-------+-------------------+
 
-            A more complex example would be, assuming we have horizon = 2, window size = 2, min window periods = 2, and we want the mean
+            A more complex example would be, assuming we have horizon = 2, window size = 2, min window size = 2, and we want the mean
```
davidbrownellWork: >window size [](start = 75, length = 11)

maxWindowSize

src/Tools/CodeGenerator/Featurizers.yaml
```diff
@@ -68,7 +68,7 @@ featurizers:
 
     configuration_params:
       - type: uint32
-        name: windowSize
+        name: maxWindowSize
```
davidbrownellWork: >name: maxWindowSize [](start = 8, length = 19)

The order of these parameters should be updated to match the order in the object's constructor.

src/Featurizers/SimpleRollingWindowFeaturizer.h
```diff
@@ -0,0 +1,359 @@
...
+
+namespace Microsoft {
+namespace Featurizer {
+namespace Featurizers {
+
+// Anonymous namespace for helper using statements
+namespace {
+    using GrainType = std::vector<std::string>;
+} // Anonymous namespace
+
```
davidbrownellWork: Give this namespace an actual name; if it remains anonymous, it will be recreated in every compilation unit.

src/Tools/CodeGenerator/Featurizers.yaml
michaelgsharp: Fixed.

src/Tools/CodeGenerator/Featurizers.yaml
michaelgsharp: Fixed.

src/Featurizers/SimpleRollingWindowFeaturizer.h
michaelgsharp: I decided to move it into the GrainFeaturizer itself as a public type.

src/Featurizers/AnalyticalRollingWindowFeaturizer.h
michaelgsharp: I may not be doing the override correctly, but when I do this it has issues because the transformer doesn't have a default constructor.

src/Featurizers/Base/RollingWindowTransformerBase.h
michaelgsharp: `flush_impl` has been added.

PR #174, opened 2020-03-19
Add in capacity() and full() for CircularBuffer

angryjinyan:


src/Featurizers/Components/WindowFeaturizerBase.h
```diff
@@ -137,6 +137,10 @@ namespace Components {
 
         size_t size() const;
 
+        size_t capacity() const;
+
+        bool full() const;
```
davidbrownellWork: >full [](start = 13, length = 4)

Can you rename this to `is_full` to communicate that it returns a `bool` value.

src/Featurizers/Components/UnitTests/WindowFeaturizerBase_UnitTest.cpp
```diff
@@ -201,10 +201,14 @@ TEST_CASE("CircularBuffer - Size 0 buffer") {
...
     NS::Featurizers::Components::CircularBuffer<std::string> circ_buf(5);
+    // capacity should be 5
+    CHECK(circ_buf.capacity() == 5);
     // size should start with 0
     CHECK(circ_buf.size() == 0);
+    // buffer is not full
+    CHECK(!circ_buf.full());
     // size would be incremented as data are pushed
     circ_buf.push("1");
     CHECK(circ_buf.size() == 1);
```
davidbrownellWork: >CHECK(circ_buf.size() == 1); [](start = 4, length = 28)

Can you add another check after this to check that capacity isn't updated even after a item is added.

src/Featurizers/Components/UnitTests/WindowFeaturizerBase_UnitTest.cpp
```diff
@@ -220,9 +224,13 @@ TEST_CASE("CircularBuffer - Size") {
     
     circ_buf.push("5");
     CHECK(circ_buf.size() == 5);
+    // buffer is full
+    CHECK(circ_buf.full());
     // size reaches the maximum of 5 and never changed
     circ_buf.push("6");
     CHECK(circ_buf.size() == 5);
```
davidbrownellWork: >CHECK(circ_buf.size() == 5); [](start = 4, length = 28)

Check that capacity remains 5 when more than 5 items have been added.

PR #176, opened 2020-03-24
Adds clear method and test for CircularBuffer

michaelgsharp:
Adds clear method and tests for CircularBuffer. This is to facilitate the flush call.


src/Featurizers/Components/WindowFeaturizerBase.h
```diff
@@ -299,6 +301,18 @@ CircularBuffer<T>::CircularBuffer(size_t max_size) : _max_size(max_size), _start
...
+    _data.clear();
+
+    // Make sure that the start offset is back to 0;
+    _start_offset = 0;
+    
+    // The c++ reference says "the vector capacity is not guaranteed to change due to calling this function."
+    // Since the capacity may change, calling reserve again just to make sure we still have the correct size.
+    _data.reserve(_max_size);
+}
+
```
angryjinyan: CircularBuffer constructor could call this method to remove some duplicate code.

michaelgsharp: Per off line discussion I have removed the duplicate code and am calling clear from the constructor.

src/Featurizers/Components/UnitTests/WindowFeaturizerBase_UnitTest.cpp
```diff
@@ -408,3 +408,48 @@ TEST_CASE("CircularBuffer - range with offset") {
     // when n and offset are not out of bound
     CHECK(++(++start_iter3) == end_iter3);
 }
+
+TEST_CASE("CircularBuffer - clear") {
```
angryjinyan: I think capacity() should also be tested to make sure capacity is still the same.

michaelgsharp: I am testing it implicitly by adding 6 items and making sure the size still stays at 5. Do you want me to add an explicit test for it as well?

angryjinyan: I think it would be clearer to have an explicit test to tell we are only resetting the size, not the capacity.

michaelgsharp: Also, capacity is implemented by returning `_maxSize` which is `const` and I can't modify it. I figured a better test instead of checking that would be to make sure the size doesn't go above what it should after I call clear. That is why I didn't have a test that actually calls `circular_buf.capacity()`

src/Featurizers/Components/UnitTests/WindowFeaturizerBase_UnitTest.cpp
```diff
@@ -408,3 +408,48 @@ TEST_CASE("CircularBuffer - range with offset") {
...
+    CHECK(circ_buf.size() == 5);
+
+    // Make sure size is correct after final clear.
+    circ_buf.clear();
+    CHECK(circ_buf.size() == 0);
+
+    // Make sure start/end iters are correct after clear
+    auto start_iter3 = circ_buf.begin();
+    auto end_iter3 = circ_buf.end();
+    CHECK(start_iter3 == end_iter3);
```
davidbrownellWork: This scenario was already verified in lines 422 - 427.

michaelgsharp: The reason I added it there again is that lines 422 to 427 don't exercise any of the circular functionality. I wanted to make sure that after the circular functionality was exercised and then clear was called, that the begin/end iterators were still correct.

PR #177, opened 2020-03-25
Preliminary support for tensors, vectors, and flush capabilities

davidbrownellWork:
Updated code gen in preparation for:

- tensors
- transformers that require flush
- enhanced vector support (used by transformers that require flush)


src/Tools/CodeGenerator/Featurizers.yaml
```diff
@@ -1244,18 +1240,17 @@ featurizers:
           - uint16
           - uint32
           - uint64
+          - bool
           - float
           - double
-          - bool
-          - string
```
michaelgsharp: Did you mean to remove string from here?

davidbrownellWork: Good catch - pending checkin.

PR #180, opened 2020-03-26
Added ForecastingPivotFeaturizer definition (status set to pending)

davidbrownellWork:


src/Tools/CodeGenerator/Featurizers.yaml
```diff
@@ -354,6 +354,107 @@ featurizers:
           - type: uint8
             name: isPaidTimeOff
 
+  # ----------------------------------------------------------------------
+  - name: ForecastingPivotFeaturizer
+    estimator_name: GrainedForecastingPivotEstimator
```
michaelgsharp: I dont think this featurizer needs to be grain aware. It has no state.
It also needs to have dynamic output.

davidbrownellWork: I agree - pending checkin.

src/Tools/CodeGenerator/Featurizers.yaml
```diff
@@ -354,6 +354,108 @@ featurizers:
...
+    num_output_columns: 1
+    description: |-
+        Similar to an Excel pivot table, this featurizer will expand values in the given output
+        where all "linked" values are not null/empty. "linked" means that all values in the same
+        column across all matrixes are not null/empty.
+
+        All rows across all matrixes must have the same length / same number of columns.
+
+        Better explained through examples, see below for a more detailed explaination of the
+        functionality.
```
angryjinyan: Would it be helpful to add to description that one input can have 0 to N output rows where N is the length/ number of columns of input matrixes?

davidbrownellWork: I'll add this when we update the status so that it is no longer pending.

PR #183, opened 2020-03-26
ForecastPivotFeaturizer

angryjinyan:


src/Featurizers/ForecastingPivotFeaturizer.h
```diff
@@ -0,0 +1,129 @@
...
+/////////////////////////////////////////////////////////////////////////
+///  \class         ForecastingPivotTransformer
+///  \brief         Expand output from upstream featurizers where all values
+///                 in the same column across matrixes are not null/empty
+///
+template <typename InputT>
+class ForecastingPivotTransformer :
+    public Components::InferenceOnlyTransformerImpl<
+        typename std::vector<Microsoft::Featurizer::RowMajMatrix<typename Microsoft::Featurizer::Traits<InputT>::nullable_type>>,
+        std::vector<typename Microsoft::Featurizer::Traits<InputT>::nullable_type>
```
davidbrownellWork: >typename Microsoft::Featurizer::Traits<InputT>::nullable_type [](start = 20, length = 61)

I think that this can be `InputT` because we will only output rows where all matrix elements are not empty.

angryjinyan: Good point! Changed to `InputT`.

src/Featurizers/ForecastingPivotFeaturizer.h
```diff
@@ -0,0 +1,129 @@
...
+    > {
+public:
+    // ----------------------------------------------------------------------
+    // |
+    // |  Public Types
+    // |
+    // ----------------------------------------------------------------------
```
davidbrownellWork: >// ---------------------------------------------------------------------- [](start = 4, length = 73)

Add a static_assert here to communicate to the caller that they don't need to pass an optional type - it will be converted to optional for them:

`static_assert(Traits<InputT>::IsNullableType == false || Traits<InputT>::IsNativeNullableType, "'InputT' should not be a nullable type");`

src/Featurizers/ForecastingPivotFeaturizer.h
```diff
@@ -0,0 +1,129 @@
...
+        std::vector<typename Microsoft::Featurizer::Traits<InputT>::nullable_type> ret;
+        Eigen::Index const col_size(input[0].cols());
+
+        for (Eigen::Index col = 0; col < col_size; ++col) {
+            bool has_nan(false);
+
+            // if any nans is found in the current column across all matrixes, this column is omitted
+            for (MatrixType const & matrix : input) {
+                // all matrixes must have the same column number
+                assert(matrix.cols() == col_size);
```
davidbrownellWork: >assert [](start = 16, length = 6)

This should be an actual error rather than an assert (which is only active for debug builds).

src/Featurizers/ForecastingPivotFeaturizer.h
```diff
@@ -0,0 +1,129 @@
...
+                    ret.emplace_back(matrix(row, col));
+                }
+
+                if (has_nan) {
+                    break;
+                }
+            }
+
+            if (!has_nan) {
+                callback(ret);
```
davidbrownellWork: >ret [](start = 25, length = 3)

`std::move(ret)`

src/Featurizers/ForecastingPivotFeaturizer.h
```diff
@@ -0,0 +1,129 @@
...
+        }
+    }
+};
+
+/////////////////////////////////////////////////////////////////////////
+///  \typedef       ForecastingPivotEstimator
+///  \brief         Estimator that creates `ForecastingPivotTransformer`.
+///
+template <typename T>
+using ForecastingPivotEstimator           = Components::InferenceOnlyEstimatorImpl<ForecastingPivotTransformer<T>>;
```
davidbrownellWork: Unfortunately, we have to give a static name to the Estimator; this means that we can't directly use the `InferenceOnlyEstimatorImpl` object. This is annoying and unintuitive, so I'd love to find a better way to pass a string in this scenario.

src/Featurizers/ForecastingPivotFeaturizer.h
michaelgsharp: @angryjinyan If you do the std::move here make sure you initialize your vector in the for loop and dont call clear at the end.

src/Featurizers/ForecastingPivotFeaturizer.h
```diff
@@ -0,0 +1,129 @@
...
+    // |
+    // ----------------------------------------------------------------------
+
+    // MSVC runs into problems when the declaration and definition are separated
+    void execute_impl(typename BaseType::InputType const &input, typename BaseType::CallbackFunction const &callback) override {
+        if (input.empty()) {
+            return;
+        }
+        
+        std::vector<typename Microsoft::Featurizer::Traits<InputT>::nullable_type> ret;
```
michaelgsharp: This should also just be InputT. We don't need the nullable type here either.

angryjinyan: Good catch! Changed to InputT.

src/Featurizers/ForecastingPivotFeaturizer.h
angryjinyan: Raising exceptions for empty input and matrixes with different column number now. Unit tests also added for these two cases.

src/Featurizers/ForecastingPivotFeaturizer.h
angryjinyan: Changed return vector scope to be within the for loop and use `std::move()`. Thank you for your advice @michaelgsharp

src/Featurizers/ForecastingPivotFeaturizer.h
angryjinyan: Gotcha! Change the `using` clause to inherit from `InferenceOnlyEstimatorImpl`.

src/Featurizers/ForecastingPivotFeaturizer.h
```diff
@@ -66,27 +68,29 @@ class ForecastingPivotTransformer :
...
+                }
 
                 for (Eigen::Index row = 0; row < matrix.rows(); ++row) {
                     if (Traits<NullableType>::IsNull(matrix(row, col))) {
                         has_nan = true;
                         break;
                     }
 
-                    ret.emplace_back(matrix(row, col));
+                    ret.emplace_back(Traits<NullableType>::GetNullableValue(matrix(row, col)));
```
michaelgsharp: Do you need to use `Traits<NullableType>::GetNullableValue` here? The vector is defined of type `InputT`, and if there is a null value we aren't outputting the row.

angryjinyan: I think `GetNullableValue` is necessary here because `nullable_type` can be nonstd::optional<T> or double_t and their `GetNullableValue` are implemented differently. In addition, I don't think it's convertible from a, say nonstd::optional<string> to string directly.

PR #186, opened 2020-03-28
Support for GrainTransformers that don't require training

davidbrownellWork:


src/Featurizers/Components/GrainFeaturizerImpl.h
```diff
@@ -215,11 +229,7 @@ class GrainEstimatorImplBase : public BaseT {
     // |  Protected Types
     // |
     // ----------------------------------------------------------------------
-    using EstimatorMap =
-        std::map<
-            GrainT,
-            EstimatorT
-        >;
+    using EstimatorMap                      = std::map<GrainT, EstimatorT>;
```
angryjinyan: Should the key type for both the estimator and transformer maps be `GrainT const &`? Or we can not use a const reference for map key.

davidbrownellWork: That's correct - we can't use references for keys as the map will outlive the backing memory associated with the references.

PR #188, opened 2020-03-30
Add in const reference

angryjinyan:
It seems like std::make_tuple does something to the qualifiers. Remove the initialization with std::make_tuple and change input type to be tuple of const references.


src/Featurizers/UnitTests/AnalyticalRollingWindowFeaturizer_UnitTest.cpp
```diff
@@ -121,10 +121,11 @@ TEST_CASE("Grained Mean - 1 grain, window size 1, horizon 1") {
     NS::AnnotationMapsPtr                   pAllColumnAnnotations(NS::CreateTestAnnotationMapsPtr(1));
     NS::Featurizers::GrainedAnalyticalRollingWindowEstimator<std::int32_t>      estimator(pAllColumnAnnotations, NS::Featurizers::AnalyticalRollingWindowCalculation::Mean, 1, 1);
 
-    using GrainedInputType = std::tuple<GrainType, std::int32_t>;
+    using GrainedInputType = std::tuple<GrainType const &, std::int32_t const &>;
```
davidbrownellWork: Change this to `using GrainedInputType = typename NS::Featurizers::GrainedAnalyticalRollingWindowEstimator<std::int32_t>::InputType;`

angryjinyan: Good point! Fixed.

PR #189, opened 2020-03-30
Support for Simple- and Analytical-RollingWindow Featurizers

davidbrownellWork:


src/Featurizers/SimpleRollingWindowFeaturizer.h
```diff
@@ -332,16 +323,33 @@ void SimpleRollingWindowEstimator<InputT, MaxNumTrainingItemsV>::complete_traini
...
+                [windowCalculation, horizon, maxWindowSize, minWindowSize](GrainType const &) {
+                    return typename SimpleRollingWindowEstimator<InputT, MaxNumTrainingItemsV>::TransformerUniquePtr(
+                        new SimpleRollingWindowTransformer<InputT, MaxNumTrainingItemsV>(
+                            windowCalculation,
+                            horizon,
+                            maxWindowSize,
+                            minWindowSize
+                        )
+                    );
+                },
```
michaelgsharp: If you are already passing in the functor to directly create the transformer and its marked as a trainingOnlyEstimator, do you even need to use the estimator directly or can you just bypass it?

michaelgsharp:
Will we need to bootstrap again after this due to the bugfix in SimpleSchema?

src/Featurizers/SimpleRollingWindowFeaturizer.h
davidbrownellWork: I had the same thought, and unfortunately the answer is that you do need the estimator creation func. This is because the GrainEstimatorImpl creates an instance of the Estimator in order to get its Name when creating the grained estimator name.

davidbrownellWork:
> Will we need to bootstrap again after this due to the bugfix in SimpleSchema?

It's possible. It turns out that I had a very old copy on my system, but you might need to do this if you find yourself rebuilding the schema.

PR #191, opened 2020-04-02
Change RollingWindow output type to matrix

angryjinyan:


src/Tools/CodeGenerator/Featurizers.yaml
```diff
@@ -1392,55 +1396,59 @@ featurizers:
...
+          # - int32
+          # - int64
+          # - uint8
+          # - uint16
+          # - uint32
+          # - uint64
+          # - string
           - float
           - double
-          - string
```
michaelgsharp, line 1438: If are making the change to only do float and double, just like the analytical rolling window, does it make sense to leave these as 2 separate featurizers? @davidbrownellWork @angryjinyan

davidbrownellWork: It does make sense to leave them as 2 different featurizers, as the limitation is due to marshaling. A hypothetical C++ framework could directly support these optional types.

michaelgsharp: Makes sense to me. I just synced with Jin ofline on this at the same time you responded as well. I hadn't thought about it from that perspective.

